PR #2, opened 2019-05-17
Energyconstruction

macumber:


lib/ladybug/energy_model/energy_construction_opaque.rb
```diff
@@ -0,0 +1,91 @@
...
+        openstudio_construction = OpenStudio::Model::Construction.new(openstudio_model)
+        openstudio_construction.setName(@hash[:name])
+        openstudio_materials = OpenStudio::Model::MaterialVector.new
+        @hash[:materials].each do |material|
+          name = material[:name]
+          material_type = material[:type] 
+          material_object = nil 
+
+          case material_type
+          when "EnergyMaterial"
```
macumber: @tanushree04 i think you want to handle all the other material types here too right?

macumber: This function would also be the place to re-implement some of the custom validator logic that does not get translated to JSON schema, like https://github.com/ladybug-tools-in2/energy-model-schema/blob/construction/app/models/energy/construction.py#L842

lib/ladybug/energy_model/energy_construction_transparent.rb
```diff
@@ -0,0 +1,94 @@
...
+          return object.get
+        end
+        return nil
+      end
+      
+      def create_openstudio_object(openstudio_model)
+        openstudio_construction = OpenStudio::Model::Construction.new(openstudio_model)
+        openstudio_construction.setName(@hash[:name])
+        openstudio_materials = OpenStudio::Model::MaterialVector.new
+        @hash[:materials].each do |material|
```
macumber: We'll want to implement custom validator logic here too.  Suggest that we override the `valid?` and `validation_errors` methods from ModelObject.  Something like:

```
def validation_errors
  # call JSON validation
  result = super()

  # implement custom validation
  if not_right_layers
    result << "Wrong layers"
  end
end
```

lib/ladybug/energy_model/energy_material.rb
```diff
@@ -0,0 +1,77 @@
...
+        if object.is_initialized
+          return object.get
+        end
+        return nil
+      end
+      
+      def create_openstudio_object(openstudio_model)
+        openstudio_opaque_material = OpenStudio::Model::StandardOpaqueMaterial.new(openstudio_model) 
+        openstudio_opaque_material.setName(@hash[:name]) 
+        openstudio_opaque_material.setRoughness(@hash[:roughness])
```
macumber: Need to handle any optional properties like this:

```
if @hash[:roughness]
  openstudio_opaque_material.setRoughness(@hash[:roughness])
else
  openstudio_opaque_material.setRoughness('Whatever the default from JSON schema is')
end
```

We want to make sure that we are providing the default values specified in the JSON schema instead of the OpenStudio defaults since they might be different.  We might see how hard it is to load the default from the schema.  Something like:

`openstudio_opaque_material.setRoughness(@schema[:EnergyMaterial][:roughness][:default])`

macumber: If there is no default in the schema then we can use the OpenStudio default.  But in that case we need to avoid passing nil in.  We should add unit tests for minimal examples of each object (e.g. having only the required fields, all optional fields excluded)

lib/ladybug/energy_model/face.rb
```diff
@@ -0,0 +1,95 @@
...
+      def find_existing_openstudio_object(openstudio_model)
+        object = openstudio_model.getSurfaceByName(@hash[:name]) #check
+        if object.is_initialized
+          return object.get
+        end
+        return nil
+      end
+      
+      def create_openstudio_object(openstudio_model)
+        openstudio_surface = OpenStudio::Model::Surface.new(openstudio_model)
```
macumber: @tanushree04 this doesn't look right, you need vertices to construct an OpenStudio Surface.  Do you have a unit test for face?

lib/ladybug/energy_model/face.rb
```diff
@@ -0,0 +1,95 @@
...
+            face_object = "Floor"
+          when "AirWall"
+            face_object = "AirWall"
+          when "Shading"
+            face_object = "Shading"
+          else
+            raise "Unknown face_type #{face_type} for face #{name}."
+        end
+        
+        openstudio_surfaces = face_object.to_openstudio(openstudio_model)
```
macumber: Are Wall, RoofCeiling, etc all different types in the JSON schema?  I thought this was like a surface type of the face?  Each JSON schema type should get a Ruby class.  Only Ruby classes have the to_openstudio method defined.

lib/ladybug/energy_model/model.rb
```diff
@@ -39,68 +41,75 @@ module Ladybug
...
+      # create openstudio objects in the openstudio model
+      def create_openstudio_objects
+        create_faces
+        create_apertures
+      end
+      
+      def create_faces
+        # TODO: create a Face class which derives from ModelObject and move all this code there
+        @hash[:faces].each do |face|
           name = face[:name]
```
macumber: This function should basically just be:

```
@hash[:faces].each do |face|
  face_object = Face.new(face)
  face_object.to_openstudio(model)
end
```

lib/ladybug/energy_model/model_object.rb
```diff
@@ -0,0 +1,114 @@
...
+        
+        @type = @hash[:type]
+        raise 'Unknown type' if @type.nil?
+        
+        @openstudio_object = nil
+      end
+      
+      # check if the ModelObject is valid
+      def valid?
+        return JSON::Validator.validate(@hash, @@schema)
```
macumber: We should change this to:

`return validation_errors.empty?`

That way we only have to override `validation_errors` in subclasses

lib/ladybug/energy_model/model_object.rb
```diff
@@ -0,0 +1,114 @@
...
+      def to_openstudio(openstudio_model)
+        
+        # return the object if we already have it
+        if @openstudio_object
+          if @openstudio_object.model == openstudio_model
+            return @openstudio_object
+          end
+        end
+        
+        @errors = []
```
macumber: Let's initialize this with `@errors = validation_errors`

lib/measures/ladybug_energy_model_measure/tests/example_face_by_face_model.json
```diff
@@ -1,15 +1,19 @@
 {
     "convert_to_meters": 1,
-    "type": "FaceByFaceModel",
+    "type": "Model",
```
macumber: @tanushree04 seems like we might want two models in the measure tests, a maximal one will all the energy model stuff defined and a minimal one without any of the energy model stuff.  We should rename the test files to reflect what they are, there is a special git rename command to rename a file once it is in git: https://help.github.com/en/articles/renaming-a-file-using-the-command-line

spec/tests/out.txt
```diff
@@ -0,0 +1,8 @@
+......[openstudio.model.StandardGlazing] <0> StandardGlazing::setWindowGlassSpectralDataSetName is deprecated, use StandardGlazing::setWindowGlassSpectralDataSet
```
macumber: @tanushree04 we should not commit this file to the repo, use git rm to remove it https://git-scm.com/docs/git-rm

spec/tests/ladybug_energy_model_model_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+    expect(extension.schema_validation_errors.empty?).to be true
+  end
+ 
+  it 'can load and validate example face by face model' do
+    file = File.join(File.dirname(__FILE__), '../files/example_face_by_face_model.json')
+    model = Ladybug::EnergyModel::Model.read_from_disk(file) 
+    expect(model.valid?).to be true
+    expect(model.validation_errors.empty?).to be true 
+    model.to_openstudio_model
+  end
```
macumber: @tanushree04 add some rspec assertions here.  For instance you should be able to call model.getFaceByName to get each face in the JSON, etc

spec/tests/ladybug_energy_model_model_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+    expect(object1).not_to be nil
+
+    material2 = Ladybug::EnergyModel::EnergyWindowMaterialShade.read_from_disk(file)
+    expect(material2.valid?).to be true
+    expect(material2.validation_errors.empty?).to be true
+    object2 = material2.to_openstudio(openstudio_model)    
+    expect(object2).not_to be nil
+    expect(object2.handle.to_s).to eq(object1.handle.to_s)
+  end
+
```
macumber: Need a unit test for Face

lib/ladybug/energy_model/energy_material.rb
tanushree04: I have set up some instance files that have only the required fields - `in_window_blind.json`, `in_window_air_gap.json`

lib/ladybug/energy_model/out.txt
```diff
@@ -0,0 +1,50 @@
+.....FFF....[openstudio.model.StandardGlazing] <0> StandardGlazing::setWindowGlassSpectralDataSetName is deprecated, use StandardGlazing::setWindowGlassSpectralDataSet
```
macumber: @tanushree04 please remove `git rm` this file

lib/ladybug/energy_model/energy_construction_opaque.rb
tanushree04: The opaque materials covered in Honeybee Primer are EnergyMaterial and EnergyMaterialNoMass which correspond to the StandardOpaqueMaterial and MasslessOpaqueMaterial respectively from the openstudio SDK

spec/tests/ladybug_energy_model_model_spec.rb
```diff
@@ -53,12 +53,14 @@
...
   end
  
+  #add assertions
   it 'can load and validate example face by face model' do
     file = File.join(File.dirname(__FILE__), '../files/example_face_by_face_model.json')
     model = Ladybug::EnergyModel::Model.read_from_disk(file) 
     expect(model.valid?).to be true
     expect(model.validation_errors.empty?).to be true 
-    model.to_openstudio_model
+    face_model = model.to_openstudio_model
```
macumber: @tanushree04 i would name this `face_openstudio_model = model.to_openstudio_model` or something to remind yourself that it is an OpenStudio Model rather than a Ladybug Model

spec/tests/ladybug_energy_model_model_spec.rb
```diff
@@ -53,12 +53,14 @@
...
+    expect(face_model.getFaceByName).not_to be nil
```
macumber: OpenStudio Model's don't have a Face class, the class name is Surface.  You could call `openstudio_model.getSurfaceByName('name')` but you have to pass in the name of the object as input

macumber: Also, `getSurfaceByName` will return an empty optional if the object does not exist in the model.  Due to Ruby quirks, this is not the same as nil.  You would have to test:

`expect(face_model.getSurfaceByName('name').empty?).not_to be true`

lib/files/schema/openapi.json
```diff
@@ -1,1419 +1,2465 @@
...
-		"securitySchemes": {
-			"JWT": {
-				"description": "JSON Web Token (JWT).\nUsage format: `Bearer <JWT>`\n",
-				"type": "http",
-				"scheme": "bearer",
-				"bearerFormat": "JWT"
-			}
-		}
-	}
+  "title": "Model",
```
macumber: This is up to date with https://github.com/ladybug-tools-in2/energy-model-schema ?

lib/ladybug/energy_model/aperture.rb
```diff
@@ -0,0 +1,90 @@
...
+
+module Ladybug
+  module EnergyModel
+    class Aperture < ModelObject
+      attr_reader :errors, :warnings
+
+      def initialize(hash)
+        super(hash)
+
+        raise "Incorrect model type '#{@type}'" unless @type == 'Aperture'
```
macumber: Should be able to check using the type from the schema right?

`raise "Incorrect model type '#{@type}'" unless @type == @hash[:type]`

macumber: Might even put that in the ModelObject initialization

lib/ladybug/energy_model/aperture.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      def create_openstudio_object(openstudio_model)
+        openstudio_vertices = OpenStudio::Point3dVector.new
+        @hash[:vertices].each do |vertex|
+          openstudio_vertices << OpenStudio::Point3d.new(vertex[:x], vertex[:y], vertex[:z])
+        end
+
+        construction = nil
+        if @hash[:energy_construction_opaque]
```
macumber: @chriswmackey we need to get on the same page about this.  When are you planning to refactor the construction schema to have a global list of constructions?

lib/ladybug/energy_model/aperture.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      def defaults
+        result = {}
+        result[:type] = @@schema[:definitions][:Aperture][:properties][:type][:enum]
+        result
+      end
+
+      def find_existing_openstudio_object(openstudio_model)
+        object = openstudio_model.getSubSurfaceByName(@hash[:name])
```
macumber: @chriswmackey are names sufficient here?  should we start using some other id?

macumber: Looks like each face has an id:

https://github.com/ladybug-tools-in2/energy-model-schema/blob/master/app/models/schemas.py#L149

lib/ladybug/energy_model/aperture.rb
```diff
@@ -0,0 +1,90 @@
...
+        else
+          construction_transparent_name = @hash[:energy_construction_transparent][:name]
+          construction_transparent = openstudio_model.getConstructionByName(construction_transparent_name)
+          unless construction_transparent.empty?
+            construction = construction_transparent.get
+          end
+        end
+
+        openstudio_subsurface = OpenStudio::Model::SubSurface.new(openstudio_vertices, openstudio_model)
+        openstudio_subsurface.setName(@hash[:name])
```
macumber: @chriswmackey we could import id, something like 

`openstudio_subsurface.additionalProperties.setFeature('LadybugId', @hash[:id])`

macumber: Seems like should we do this on all objects?  We can add code to build an index of OpenStudio objects by Ladybug Id.

lib/ladybug/energy_model/aperture.rb
macumber: Should the face refer to a construction by name or id?

lib/ladybug/energy_model/model_object.rb
```diff
@@ -0,0 +1,130 @@
...
+        @openstudio_object = nil
+      end
+
+      def defaults
+        raise 'defaults not implemented for ModelObject, override in your class'
+      end
+
+      # check if the ModelObject is valid
+      def valid?
+        JSON::Validator.validate(@hash, @@schema)
```
macumber: Let's change this to:

`return validation_errors.empty?`

That way custom validation rules in overridden `validation_errors` methods will also be considered

lib/ladybug/energy_model/energy_material.rb
```diff
@@ -0,0 +1,136 @@
...
+          openstudio_opaque_material.setRoughness(@@schema[:definitions][:EnergyMaterial][:properties][:roughness][:default])
+        end
+        openstudio_opaque_material.setThickness(@hash[:thickness])
+        openstudio_opaque_material.setConductivity(@hash[:conductivity])
+        openstudio_opaque_material.setDensity(@hash[:density])
+        openstudio_opaque_material.setSpecificHeat(@hash[:specific_heat])
+        if @hash[:thermal_absorptance]
+          openstudio_opaque_material.setThermalAbsorptance(@hash[:thermal_absorptance].to_f)
+        else
+          openstudio_opaque_material.setThermalAbsorptance(@@schema[:definitions][:EnergyMaterial][:properties][:thermal_absorptance][:default].to_f)
```
macumber: This case can't happen because the default value is already set in the constructor right?

macumber: I guess someone could call the constructor with a hash setting thermal_absorptance to nil

lib/ladybug/energy_model/energy_material.rb
```diff
@@ -0,0 +1,136 @@
...
+          openstudio_opaque_material.setRoughness(@hash[:roughness])
+        else
+          openstudio_opaque_material.setRoughness(@@schema[:definitions][:EnergyMaterial][:properties][:roughness][:default])
+        end
+        openstudio_opaque_material.setThickness(@hash[:thickness])
+        openstudio_opaque_material.setConductivity(@hash[:conductivity])
+        openstudio_opaque_material.setDensity(@hash[:density])
+        openstudio_opaque_material.setSpecificHeat(@hash[:specific_heat])
+        if @hash[:thermal_absorptance]
+          openstudio_opaque_material.setThermalAbsorptance(@hash[:thermal_absorptance].to_f)
```
macumber: Why does this need a to_f but other setters don't?

lib/ladybug/energy_model/energy_material.rb
```diff
@@ -0,0 +1,136 @@
...
+        result = {}
+        result[:type] = 'EnergyMaterial'
+        result[:roughness] = @@schema[:definitions][:EnergyMaterial][:properties][:roughness][:default]
+        result[:thermal_absorptance] = @@schema[:definitions][:EnergyMaterial][:properties][:thermal_absorptance][:default]
+        result[:solar_absorptance] = @@schema[:definitions][:EnergyMaterial][:properties][:solar_absorptance][:default]
+        result[:visible_absorptance] = @@schema[:definitions][:EnergyMaterial][:properties][:visible_absorptance][:default]
+        result
+      end
+
+      def name
```
macumber: Do we need these methods anymore?

lib/ladybug/energy_model/face.rb
```diff
@@ -0,0 +1,103 @@
...
+        nil
+      end
+
+      def create_openstudio_object(openstudio_model)
+        openstudio_vertices = OpenStudio::Point3dVector.new
+        @hash[:vertices].each do |vertex|
+          openstudio_vertices << OpenStudio::Point3d.new(vertex[:x], vertex[:y], vertex[:z])
+        end
+
+        parent_name = @hash[:parent][:name]
```
macumber: @chriswmackey should this also create a ThermalZone?

lib/ladybug/energy_model/model.rb
```diff
@@ -39,116 +43,83 @@ module Ladybug
...
+        @type = @hash[:type]
+        raise 'Unknown model type' if @type.nil?
+        raise "Incorrect model type '#{@type}'" unless @type == 'Model'
       end
-      
+
       # check if the model is valid
       def valid?
-        return JSON::Validator.validate(@model, @@schema, {:fragment => "#/components/schemas/#{@model_type}"})
+        JSON::Validator.validate(@hash, @@schema)
```
macumber: Change this to 

`return validation_errors.empty?`

lib/ladybug/energy_model/model.rb
```diff
@@ -39,116 +43,83 @@ module Ladybug
...
       def validation_errors
-        return JSON::Validator.fully_validate(@model, @@schema, {:fragment => "#/components/schemas/#{@model_type}"})
-      end
-      
-      # convert to a new openstudio model
-      def to_openstudio
-        osm = OpenStudio::Model::Model.new
-        create_openstudio_objects(osm)
-        return osm
+        JSON::Validator.fully_validate(@hash, @@schema)
```
macumber: This won't run custom validation checks, that ok?

lib/ladybug/energy_model/model_object.rb
```diff
@@ -0,0 +1,130 @@
...
+      # convert ModelObject to an openstudio object
+      def to_openstudio(openstudio_model)
+        # return the object if we already have it
+        if @openstudio_object
+          if @openstudio_object.model == openstudio_model
+            return @openstudio_object
+          end
+        end
+
+        @errors = validation_errors
```
macumber: We want to bubble these up to the top so we can see all errors and warnings after a full model translation, this is a todo for @macumber

lib/measures/ladybug_energy_model_measure/measure.rb
```diff
@@ -71,33 +70,33 @@ def arguments(model)
   # define what happens when the measure is run
   def run(model, runner, user_arguments)
     super(model, runner, user_arguments)
-puts "hello!"
-STDOUT.flush
+    puts 'hello!'
```
macumber: Should get rid of puts statements like this before merging

lib/files/schema/openapi.json
tanushree04: @macumber  yes it is

lib/ladybug/energy_model/model.rb
tanushree04: I can add the custom validations in the schema to the ruby code.

macumber: Well the issue is that we want to get the validation errors from each object, let me think about this one https://github.com/ladybug-tools-in2/energy-model-measure/issues/3

lib/ladybug/energy_model/model_object.rb
macumber: https://github.com/ladybug-tools-in2/energy-model-measure/issues/3

lib/ladybug/energy_model/energy_material.rb
tanushree04: The default value would be taken from the schema in case no default is specified in the `openstudio_model` since its not a required property.

lib/ladybug/energy_model/schedule_ruleset.rb
```diff
@@ -0,0 +1,134 @@
...
+        schedule_type_limits.setUpperLimitValue(@hash[:schedule_type_limits][:upper_limit_value])
+        case 
+        when @hash[:schedule_type_limits][:numeric_type][:type] == 'ScheduleContinuous'
+          schedule_type_limits.setNumericType('Continuous')
+        when @hash[:schedule_type_limits][:numeric_type][:type] == 'ScheduleDiscrete'
+          schedule_type_limits.setNumericType('Discrete')
+        end
+        openstudio_summer_day = OpenStudio::Model::ScheduleDay.new(openstudio_model)
+        openstudio_schedule_ruleset.setSummerDesignDaySchedule(openstudio_summer_day)
+        openstudio_schedule_ruleset.summerDesignDaySchedule.setName(@hash[:summer_designday_schedule][:name])
```
macumber: Would save some line space to just call `openstudio_summer_day.setName(@hash[:summer_designday_schedule][:name])`, the objects are linked so you don't have to get it from the openstudio_schedule_ruleset each time

lib/ladybug/energy_model/schedule_ruleset.rb
```diff
@@ -0,0 +1,134 @@
...
+        if @hash[:summer_designday_schedule][:interpolate_to_timestep]
+          openstudio_schedule_ruleset.summerDesignDaySchedule.setInterpolatetoTimestep(@hash[:summer_designday_schedule][:interpolate_to_timestep])
+        else 
+          openstudio_schedule_ruleset.summerDesignDaySchedule.setInterpolatetoTimestep(@@schema[:definitions][:ScheduleDay][:properties][:interpolate_to_timestep][:default])
+        end
+        @hash[:summer_designday_schedule][:day_values].each do |day_values|
+          openstudio_schedule_ruleset.summerDesignDaySchedule.addValue(OpenStudio::Time.new(0, day_values[:time][:hour], day_values[:time][:minute],0),day_values[:value_until_time])
```
macumber: Need to make sure that midnight is handled correctly, you will want a test or two for that

macumber: Also, what if last time is 10pm or something before midnight, what value is used for rest of the day?

lib/ladybug/energy_model/schedule_ruleset.rb
```diff
@@ -0,0 +1,134 @@
...
+        schedule_type_limits.setUnitType(@hash[:schedule_type_limits][:unit_type])
+        schedule_type_limits.setLowerLimitValue(@hash[:schedule_type_limits][:lower_limit_value])
+        schedule_type_limits.setUpperLimitValue(@hash[:schedule_type_limits][:upper_limit_value])
+        case 
+        when @hash[:schedule_type_limits][:numeric_type][:type] == 'ScheduleContinuous'
+          schedule_type_limits.setNumericType('Continuous')
+        when @hash[:schedule_type_limits][:numeric_type][:type] == 'ScheduleDiscrete'
+          schedule_type_limits.setNumericType('Discrete')
+        end
+        openstudio_summer_day = OpenStudio::Model::ScheduleDay.new(openstudio_model)
```
macumber: You could probably factor out the code to make a schedule day from a hash object to a new function and avoid duplicating code for summer, winter, default, and the rest of the schedule days for each rule

spec/files/schedule_ruleset.json
```diff
@@ -1 +1 @@
...
...le": {"type": "ScheduleDay", "name": "Default Day 1", "interpolate_to_timestep": false, "day_values": [{"time": {"hour": 24, "minute": 60}, "value_until_time": 20}]}, "summer_designday_schedule": {"type": "ScheduleDay", "name": "Default Day 2", "interpolate_to_timestep": false, "day_values": [{"time": {"hour": 24, "minute": 60}, "value_until_time": 20}]}, "winter_designday_schedule": {"type": "ScheduleDay", "name": "Default Day 3", "interpolate_to_timestep": false, "day_values": [{"time": {"hour": 24, "minute": 60}, "value_until_time": 20}]}, "schedule_rules": [{"type": "ScheduleRule", "schedule_day": {"type": "ScheduleDay", "name": "Default Day 4", "interpolate_to_timestep": false, "day_values": [{"time": {"hour": 24, "minute": 60}, "value_until_time": 20}]}, "start_period": {"date": {"month": 1, "day": 1}, "time": {"hour": 0, "minute": 0}, "is_leap_year": false}, "end_period": {"date": {"month": 12, "day": 31}, "time": {"hour": 24, "minute": 60}, "is_leap_year": false}, "name": "Schedule Rule 1", "apply_sunday": true, "apply_monday": true, "apply_tuesday": true, "apply_wednesday": true, "apply_thursday": true, "apply_friday": true, "apply_saturday": true, "apply_holiday": true}]}
```
macumber: I don't think hour = 24, minute = 60 is a valid time, The end of the day would be hour = 24, minute =60

spec/tests/ladybug_energy_schedule_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  it 'can load and validate example schedule ruleset 1' do 
+    openstudio_model = OpenStudio::Model::Model.new
+    file = File.join(File.dirname(__FILE__), '../files/schedule_ruleset_1.json')
+    model = Ladybug::EnergyModel::ScheduleRuleset.read_from_disk(file)
+    expect(model.valid?).to be true
+    expect(model.validation_errors.empty?).to be true
+    openstudio_model = model.to_openstudio(openstudio_model)
+    expect(openstudio_model).not_to be nil
+  end
```
macumber: We should add more assertions that all the schedules are in the openstudio model, they have the right day schedules, with the right values, etc.

spec/tests/ladybug_energy_schedule_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  it 'has a valid schema' do
+    extension = Ladybug::EnergyModel::Extension.new
+    expect(extension.schema.nil?).to be false
+    expect(extension.schema_valid?).to be true
+    expect(extension.schema_validation_errors.empty?).to be true
+  end
+
+  it 'can load and validate example schedule ruleset' do
```
macumber: Should also add a test to load a JSON with an invalid schedule, test what happens.

spec/tests/ladybug_energy_schedule_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    expect(File.exist?(extension.files_dir)).to be true
+  end
+
+  it 'has a valid schema' do
+    extension = Ladybug::EnergyModel::Extension.new
+    expect(extension.schema.nil?).to be false
+    expect(extension.schema_valid?).to be true
+    expect(extension.schema_validation_errors.empty?).to be true
+  end
+
```
macumber: To make testing easier you can also create a schedule_ruleset JSON object directly in the test and call the ScheduleRuleset constructor with it.  Its a good idea to test the full model conversion too, but also a good idea to test the object directly.

lib/ladybug/energy_model/model.rb
```diff
@@ -68,7 +68,7 @@ def initialize(hash)
 
       # check if the model is valid
       def valid?
-        JSON::Validator.validate(@hash, @@schema)
+        return validation_errors.empty?
       end
 
       # return detailed model validation errors
```
macumber: I think we will need to add a new method like `create_schedules` and call it in `create_openstudio_objects` to actually import the schedules from the JSON

tanushree04: #4

lib/ladybug/energy_model/schedule_ruleset.rb
```diff
@@ -0,0 +1,134 @@
...
+          if schedule_rule[:schedule_day][:interpolate_to_timestep]
+            openstudio_schedule_rule.daySchedule.setInterpolatetoTimestep(schedule_rule[:schedule_day][:interpolate_to_timestep])
+          else
+            openstudio_schedule_rule.daySchedule.setInterpolatetoTimestep(@@schema[:definitions][:ScheduleDay][:properties][:interpolate_to_timestep][:default])
+          end
+          schedule_rule[:schedule_day][:day_values].each do |day_values|
+            openstudio_schedule_rule.daySchedule.addValue(OpenStudio::Time.new(0,day_values[:time][:hour], day_values[:time][:minute],0), day_values[:value_until_time])
+          end
+          openstudio_start_date = openstudio_schedule_rule.setStartDate(OpenStudio::Date.new(OpenStudio::MonthOfYear.new(schedule_rule[:start_period][:date][:month]), schedule_rule[:start_period][:date][:day]))
+          #setStartTime?
```
tanushree04: Add Start Time.

lib/ladybug/energy_model/schedule_ruleset.rb
tanushree04: Should it raise an error if the last day_values[:time][:hour] is not equal to 24? Since OS also requires values to be specified for a 24-hour period.

spec/files/schedule_ruleset.json
tanushree04: I will update the maximum limit for minutes to 59 in the schema.

PR #9, opened 2019-10-16
feat(bug_fixes): Construction measure

### LINKED ISSUE #7: Hard-assign constructions for Adiabatic Faces
We know that we want the adiabatic constructions to follow the same exact ConstructionSet logic as the interior constructions but the OpenStudio DefaultConstrucitonSet object does not currently support this.  Therefore, we need a script that a script that runs at the end of the model creation. It would just walk through all of the Surfaces, find any Adiabatic ones, check the ConstructionSet governing the Surface, and hard-assign the Interior construction for that Surface type.

### LINKED ISSUE #8: Avoid creating a separate ShadingSurfaceGroup for each and every Shade
Right now, if I have a set of louvers assigned to an Aperture, each and every louver gets its own shading group, which is not at all efficient and I don't think it the way that shading groups were intended to be used. Instead, shades should be grouped according to the Aperture, Face or Room to which they belong. To do this, an edit needs to be made to the `Shade.create_openstudio_object` method such that it tries to first get the existing shading group from the parent object and add the shade to that group before it goes to try and create a new shading object.

tanushree04:
- Hard assigning construction for adiabatic surfaces.
- Creating ShadingSurfaceGroup at room, face and aperture level, creating separate ShadingSurfaceGroup for orphaned_shades for building. 
- Assigning adjacent surfaces at model level by looping through the surface JSONs. (TODO: setAdjacentSurface failing)

This closes #8, closes #7

### DIFF
--- a/lib/ladybug/energy_model/face.rb
+++ b/lib/ladybug/energy_model/face.rb
@@ -77,10 +77,8 @@ module Ladybug
         openstudio_surface.setName(@hash[:name])
         openstudio_surface.setSurfaceType(@hash[:face_type])
         openstudio_surface.setConstruction(openstudio_construction) if openstudio_construction
-        openstudio_surface.setOutsideBoundaryCondition(@hash[:boundary_condition][:type])
         boundary_condition = (@hash[:boundary_condition][:type])
        
-
         case boundary_condition
         when 'Outdoors'
           if @hash[:boundary_condition][:sun_exposure] == true
@@ -99,16 +97,26 @@ module Ladybug
             openstudio_surface.setViewFactortoGround(@hash[:boundary_condition][:view_factor])
           end
         when 'Surface'
-          surface = nil 
           if @hash[:boundary_condition][:boundary_condition_objects][0]
+            surface = nil
             surface_object = openstudio_model.getSurfaceByName(@hash[:boundary_condition][:boundary_condition_objects][0])
             unless surface_object.empty?
               surface = surface_object.get
               openstudio_surface.setAdjacentSurface(surface)
             end
           end
+                   
+          #key = @hash
+          #value = @hash[:boundary_condition][:boundary_condition_objects][0]
+          
+          #$surfaces.store(:key, value)
+
+          #$surfaces.merge!(key: value)
+
+          puts "#{$surfaces}"
         end
 
+        openstudio_surface.setOutsideBoundaryCondition(@hash[:boundary_condition][:type]) unless @hash[:boundary_condition][:type] == 'Surface'
 
         if @hash[:apertures]
           @hash[:apertures].each do |aperture|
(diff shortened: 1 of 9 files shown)


lib/ladybug/energy_model/face.rb
```diff
@@ -99,16 +97,26 @@ def create_openstudio_object(openstudio_model)
             openstudio_surface.setViewFactortoGround(@hash[:boundary_condition][:view_factor])
           end
         when 'Surface'
-          surface = nil 
           if @hash[:boundary_condition][:boundary_condition_objects][0]
+            surface = nil
             surface_object = openstudio_model.getSurfaceByName(@hash[:boundary_condition][:boundary_condition_objects][0])
             unless surface_object.empty?
               surface = surface_object.get
               openstudio_surface.setAdjacentSurface(surface)
```
chriswmackey: I just realized that, while this original code here won't set the adjacent surface correctly the first time around (ie. when adding surface1 to a Model, which is adjacent to surface2), it will set the adjacency correctly on the second time around (when adding surface2 to the model and setting it to be adjacent to surface1, which should already be in the model by that point). So I think the original code here might actually be ok and I just wasn't thinking correctly about it before. @tanushree04 , in your tests so far, can you see whether the adjacent surfaces of the `model_multi_zone_office.json` model are being set correctly? If so, we can just keep this code the way it is (without the need for the global hash) and you can also get rid of the commented code that you have below here.

lib/ladybug/energy_model/model.rb
```diff
@@ -201,19 +201,46 @@ def create_rooms
...
+
+              openstudio_adjacent_surface = nil
+              openstudio_adjacent_surface_object = @openstudio_model.getSurfaceByName(adjacent_surface_name)
+              unless openstudio_adjacent_surface_object.empty?
+                openstudio_adjacent_surface = openstudio_adjacent_surface_object.get
+              end
+
+              #openstudio_surface.setAdjacentSurface(openstudio_adjacent_surface)
+              end
+            end
```
chriswmackey: See my comment in the `Face` class since I think we may be able to get rid of this.

lib/ladybug/energy_model/room.rb
```diff
@@ -77,9 +77,39 @@ def create_openstudio_object(openstudio_model)
           face = Face.new(face)
           openstudio_face = face.to_openstudio(openstudio_model)
           openstudio_face.setSpace(openstudio_space)
+
           nil
         end
 
+        openstudio_surfaces = openstudio_space.surfaces.each do |surface|
+          if surface.outsideBoundaryCondition == 'Adiabatic'
```
chriswmackey: I think you need an extra check here. You only want to do this in the event that the surface boundary condition is adiabatic AND there isn't already a construction hard-assigned to the surface. Otherwise, you might be overwriting a construction that is intended to be hard-assigned here from the honeybee schema.

lib/ladybug/energy_model/face.rb
tanushree04: @chriswmackey you're right, I ran the measure and the surfaces are correctly assigned as adjacent surfaces in the osm using the earlier code.

chriswmackey: Glad to hear that is the case. Let's just use your original solution, then. I see you commented out the code for the new one we had been considering with the global hash and I would say that we can just delete that code rather than commenting it out (better to keep the code clean). Once you do that, feel free to merge this PR in!

tanushree04:
@chriswmackey This is ready to merge.

PR #10, opened 2019-10-28
feat(schedules, spacetype): Added ScheduleFixedInterval, ScheduleRuleset, ScheduleTypeLimits, SpaceTypes

tanushree04:
This PR includes schedules and space types.

Issues to address: 

- [ ] Add check within ScheduleRuleset to see if there are duplicate ScheduleDay and only convert one to OSM. 
- [x] Translation of latent_fraction for People within SpaceType.
- [x] Translation of flow_per_zone for Ventilation within SpaceType.
- [x] Add more specific tests to check translation of schedules.


lib/ladybug/energy_model/schedule_ruleset_abridged.rb
```diff
@@ -0,0 +1,134 @@
...
+            openstudio_schedule_rule.setApplySunday(rule[:apply_sunday])
+            openstudio_schedule_rule.setApplyMonday(rule[:apply_monday])
+            openstudio_schedule_rule.setApplyTuesday(rule[:apply_tuesday])
+            openstudio_schedule_rule.setApplyWednesday(rule[:apply_wednesday])
+            openstudio_schedule_rule.setApplyThursday(rule[:apply_thursday])
+            openstudio_schedule_rule.setApplyFriday(rule[:apply_friday])
+            openstudio_schedule_rule.setApplySaturday(rule[:apply_saturday])
+            year_description = openstudio_model.getYearDescription
+            start_date = year_description.makeDate(rule[:start_date][0], rule[:start_date][1])         
+            end_date = year_description.makeDate(rule[:end_date][0], rule[:end_date][1])
```
chriswmackey: Again, I am a little confused why you need the year description here.

lib/ladybug/energy_model/schedule_ruleset_abridged.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        openstudio_schedule_ruleset.setSummerDesignDaySchedule(summer_designday_schedule)
+
+        if @hash[:winter_designday_schedule]
+          winter_designday_schedule = OpenStudio::Model::ScheduleDay.new(openstudio_model)
+          values = @hash[:winter_designday_schedule][:values]
+          times = @hash[:winter_designday_schedule][:times]
+          values.each_index do |i|
+            winter_designday_schedule.addValue(OpenStudio::Time.new(0,times[i][0],times[i][1] ,0), values[i])
+          end
```
chriswmackey: Given that you have these 5 lines repeated 3 times in this module, it might be a bit cleaner to just wrap them into their own hidden function and just pass the times and values to it.

lib/ladybug/energy_model/schedule_ruleset_abridged.rb
```diff
@@ -0,0 +1,134 @@
...
+          schedule_type_limit = openstudio_model.getScheduleTypeLimitsByName(@hash[:schedule_type_limit])
+          unless schedule_type_limit.empty?
+            schedule_type_limit_object = schedule_type_limit.get
+            openstudio_schedule_ruleset.setScheduleTypeLimits(schedule_type_limit_object)
+          end
+        end
+
+        if @hash[:schedule_rules]
+          @hash[:schedule_rules].each do |rule|          
+            openstudio_schedule_rule = OpenStudio::Model::ScheduleRule.new(openstudio_schedule_ruleset)
```
chriswmackey: You might have already checked this but do you know for sure if this adds the rules in the right order?  In the Honeybee json, they are ordered from highest priority to lowest priority and I am not sure if this way of initializing ScheduleRules in the SDK preserves this order.

lib/ladybug/energy_model/space_type.rb
```diff
@@ -0,0 +1,284 @@
...
+          openstudio_people_definition.setName(@hash[:people][:name])
+          openstudio_people_definition.setPeopleperSpaceFloorArea(@hash[:people][:people_per_area])
+          if @hash[:people][:radiant_fraction]
+            openstudio_people_definition.setFractionRadiant(@hash[:people][:radiant_fraction])
+          else
+            openstudio_people_definition.setFractionRadiant(@@schema[:definitions][:PeopleAbridged][:radiant_fraction][:default])
+          end
+          #TODO if @hash[:people][:latent_fraction]
+          
+          openstudio_people = OpenStudio::Model::People.new(openstudio_people_definition)
```
chriswmackey: I think you will want to pull out the creation of People, Lights, Equipment, etc. into their own modules. The way you have it now, this code is going to be duplicated for when you have a Room with a custom People definition that overrides the the Space Type.

lib/ladybug/energy_model/space_type.rb
```diff
@@ -0,0 +1,284 @@
...
+          heating_schedule_object = heating_schedule.get
+        end
+        openstudio_thermostat.setHeatingSetpointTemperatureSchedule(heating_schedule_object)
+        
+        cooling_schedule_object = nil
+        cooling_schedule = openstudio_model.getScheduleByName(@hash[:setpoint][:cooling_schedule])
+        unless cooling_schedule.empty?
+          cooling_schedule_object = cooling_schedule.get
+        end
+        openstudio_thermostat.setCoolingSetpointTemperatureSchedule(cooling_schedule_object)
```
chriswmackey: Yea, you definitely want this to be it's own module. You want to create the thermostat when you have a zone to apply it to. Doing it for the whole space type is not going to make it applicable to all of the Zones.

lib/ladybug/energy_model/space_type.rb
tanushree04: @chriswmackey I realise what I did. I am assigning the space types but not creating the individual modules for People, Lights, etc at the room level.  I had that in the schema correctly, but have not implemented in the measure.
I will update this for the measure as well.

lib/ladybug/energy_model/schedule_fixed_interval_abridged.rb
```diff
@@ -0,0 +1,99 @@
...
+          openstudio_schedule_fixed_interval.setIntervalLength(60/timestep)
+        else
+          timestep = @@schema[:definitions][:ScheduleFixedIntervalAbridged][:properties][:timestep][:default]
+          interval_length = 60/timestep
+          openstudio_schedule_fixed_interval.setIntervalLength(60/timestep)
+        end
+        openstudio_interval_length = OpenStudio::Time.new(0,0,interval_length) 
+
+        year_description = openstudio_model.getYearDescription
+        start_date = year_description.makeDate(@hash[:start_date][:month], @hash[:start_date][:day])
```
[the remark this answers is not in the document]
tanushree04: @chriswmackey I spoke to @DavidGoldwasser about this, OpenStudio requires the year to create a date object. The `getYearDescription` method either gets the year from the model or creates a year object if it doesn't exist.

lib/ladybug/energy_model/schedule_ruleset_abridged.rb
tanushree04: @chriswmackey I dont think I will be reuse the function for defaultDaySchedule since I don't create a new ScheduleDay for it but add values to schedule_ruleset.defaultDaySchedule. 
I have clubbed these lines in a new function for summer and winter design day however.

lib/ladybug/energy_model/gas_equipment_abridged.rb
```diff
@@ -0,0 +1,91 @@
...
+      def create_openstudio_object(openstudio_model)
+        openstudio_gas_equipment_definition = OpenStudio::Model::GasEquipmentDefinition.new(openstudio_model)
+        openstudio_gas_equipment_definition.setWattsperSpaceFloorArea(@hash[:watts_per_area])
+        if @hash[:gas_equipment][:radiant_fraction]
+          openstudio_gas_equipment_definition.setFractionRadiant(@hash[:radiant_fraction])
+        else 
+          openstudio_gas_equipment_definition.setFractionRadiant(@@schema[:definitions][:GasEquipmentAbridged][:properties][:radiant_fraction][:default])
+        end
+        if @hash[:gas_equipment][:latent_fraction]
+          openstudio_electric_equipment_definition.setFractionLatent(@hash[:gas_equipment][:latent_fraction])
```
chriswmackey: You have a typo here, which is pretty clearly a relic of some copy/pasting.  Maybe this is an indication that both Gas and Electric Equipment should be inheriting from a common class so that you end up with less duplicated code between (this is what I have done on the Python end). Maybe it's a bit much for this PR but you need to at least look through all of the duplicate code for typos, then.

lib/ladybug/energy_model/gas_equipment_abridged.rb
```diff
@@ -0,0 +1,91 @@
...
+        else
+          openstudio_lights_definition.setFractionLatent(@@schema[:definitions][:GasEquipmentAbridged][:properties][:latent_fraction][:default])
```
chriswmackey: Another typo.

lib/ladybug/energy_model/electrical_equipment_abridged.rb
```diff
@@ -0,0 +1,90 @@
...
+        openstudio_electric_equipment_definition.setWattsperSpaceFloorArea(@hash[:watts_per_area])
+        if @hash[:radiant_fraction]
+          openstudio_electric_equipment_definition.setFractionRadiant(@hash[:radiant_fraction])
+        else 
+          openstudio_electric_equipment_definition.setFractionRadiant(@@schema[:definitions][:ElectricalEquipmentAbridged][:properties][:radiant_fraction][:default])
+        end
+        if @hash[:latent_fraction]
+          openstudio_electric_equipment_definition.setFractionLatent(@hash[:latent_fraction])
+        else
+          openstudio_lights_definition.setFractionLatent(@@schema[:definitions][:ElectricalEquipmentAbridged][:properties][:latent_fraction][:default])
```
chriswmackey: typo

lib/ladybug/energy_model/electrical_equipment_abridged.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+        if @hash[:lost_fraction]
+          openstudio_lights_definition.setFractionLost(@hash[:lost_fraction])
```
chriswmackey: typo

lib/ladybug/energy_model/electrical_equipment_abridged.rb
```diff
@@ -0,0 +1,90 @@
...
+        else 
+          openstudio_lights_definition.setFractionLatent(@@schema[:definitions][:ElectricalEquipmentAbridged][:properties][:lost_fraction][:default])
```
chriswmackey: typo

lib/ladybug/energy_model/infiltration_abridged.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    
+      def defaults
+        result = {}
+        result
+      end
+    
+      def find_existing_openstudio_object(openstudio_model)
+        model_gas_equipment = openstudio_model.getSpaceInfiltrationDesignFlowRateByName(@hash[:name])
+        return model_gas_equipment.get unless model_gas_equipment.empty?
```
chriswmackey: 2 typos here

tanushree04:
@chriswmackey this latest commit includes the humidistat and thermostat objects. Currently the name for these objects is the same as the SetpointAbridged name.

lib/ladybug/energy_model/schedule_ruleset_abridged.rb
tanushree04: @chriswmackey I've added code to assign the `ScheduleIndex` to the ScheduleRules. In OpenStudio, it follows the same prority 0 is the highest priority rule.

lib/ladybug/energy_model/people_abridged.rb
```diff
@@ -0,0 +1,89 @@
...
+    
+      def create_openstudio_object(openstudio_model)
+        openstudio_people_definition = OpenStudio::Model::PeopleDefinition.new(openstudio_model)
+        openstudio_people_definition.setPeopleperSpaceFloorArea(@hash[:people_per_area])
+        if @hash[:radiant_fraction]
+          openstudio_people_definition.setFractionRadiant(@hash[:radiant_fraction])
+        else
+          openstudio_people_definition.setFractionRadiant(@@schema[:definitions][:PeopleAbridged][:radiant_fraction][:default])
+        end
+        #TODO if @hash[:people][:latent_fraction]
```
chriswmackey: I can see that the OpenStudio SDK let's you set the People `SensibleHeatFraction`, which is just equal to 1 - `latent_fraction`. So you should use this to make sure this variable included when it's not the default `autocalculate`.  Here's the relevant part of the SDK documentation:
https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-2.9.0-doc/model/html/classopenstudio_1_1model_1_1_people_definition.html

lib/ladybug/energy_model/room.rb
```diff
@@ -121,7 +119,86 @@ def create_openstudio_object(openstudio_model)
...
+        #  end
+        #end
+
+        #if @hash[:properties][:energy][:setpoint][:humidification_schedule] or @hash[:properties][:energy][:setpoint][:dehumidification_schedule]
+        #  setpoint_humidistat_object = nil
+        #  setpoint_humidistat = openstudio_model.getZoneControlHumidistatByName(@hash[:properties][:energy][:setpoint])
+        #  unless setpoint_humidistat_object.empty?
+        #    setpoint_humidistat_object = setpoint_humidistat.get
+        #  end
+        #end
```
chriswmackey: Is this code commented out here because you plan to send a separate PR that adds a thermal zone to each space?  Or are you still planning to address it as part of this PR?

chriswmackey: If it's the former, putting a `TODO` comment before all of these comments is good practice.

lib/ladybug/energy_model/schedule_fixed_interval_abridged.rb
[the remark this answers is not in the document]
chriswmackey: Ok. I know the `YearDescription` is different in C# bindings vs Ruby so maybe that's the reason why it seemed odd to me. Let's keep it as-is.

lib/ladybug/energy_model/setpoint_humidistat_abridged.rb
```diff
@@ -0,0 +1,83 @@
...
+        raise "Incorrect model type '#{@type}'" unless @type == 'SetpointAbridged'
+      end
+    
+      def defaults
+        result = {}
+        result
+      end
+    
+      def find_existing_openstudio_object(openstudio_model)
+        model_setpoint_humidistat = openstudio_model.getZoneControlHumistatByName(@hash[:name])
```
chriswmackey: You really shouldn't have a case where you use a method like this since each Zone/Space needs to get its own humidistat (it's not like people or lighting where you can use the same definition all over the place).

lib/ladybug/energy_model/setpoint_thermostat_abridged.rb
```diff
@@ -0,0 +1,79 @@
...
+        raise "Incorrect model type '#{@type}'" unless @type == 'SetpointAbridged'
+      end
+    
+      def defaults
+        result = {}
+        result
+      end
+    
+      def find_existing_openstudio_object(openstudio_model)
+        model_setpoint_thermostat = openstudio_model.getThermostatSetpointDualSetpointByName(@hash[:name])
```
chriswmackey: Same thing here. There should be only one thermostat per Zone/Space.  Also, I am not sure if you need to name these modules for thermostat and humidistat modules with `abridged` in the title. Sure, the honeybee `Setpoint` object is abridged but there's no Honeybee thermostat object so the word `abridged` here seems a bit misplaced.

lib/ladybug/energy_model/ventilation_abridged.rb
```diff
@@ -0,0 +1,89 @@
...
+    
+      def create_openstudio_object(openstudio_model)       
+        openstudio_ventilation = OpenStudio::Model::DesignSpecificationOutdoorAir.new(openstudio_model)
+         openstudio_ventilation.setName(@hash[:name])
+         if @hash[:air_changes_per_hour]
+           openstudio_ventilation.setOutdoorAirFlowAirChangesperHour(@hash[:air_changes_per_hour])
+         else
+           openstudio_ventilation.setOutdoorAirFlowAirChangesperHour(@@schema[:definitions][:VentilationAbridged][:properties][:air_changes_per_hour][:default])
+         end
+         #TODO if @hash[:ventilation][:flow_per_zone]
```
chriswmackey: I see there's a method to `setOutdoorAirFlowRate` on the `DesignSpecificationOutdoorAir` object in the SDK:
https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-2.9.0-doc/model/html/classopenstudio_1_1model_1_1_design_specification_outdoor_air.html
This is the method that you should be using to assign this variable.

tanushree04:
@chriswmackey the latest commit includes measure for IdealAirSystem.

lib/ladybug/energy_model/room.rb
```diff
@@ -138,6 +138,10 @@ def create_openstudio_object(openstudio_model)
           unless people.empty?
             people_object = people.get
             people_object.setSpace(openstudio_space)
+          else
```
tanushree04: @chriswmackey I added an else statement to create PeopleAbridged in case it doesn't exist.

lib/ladybug/energy_model/ideal_air_system.rb
```diff
@@ -0,0 +1,99 @@
...
+          openstudio_ideal_air.setSensibleHeatRecoveryEffectiveness(@@schema[:definitions][:IdealAirSystem][:properties][:sensible_heat_recovery][:default]) #TODO : Openstudio defaults are different from schema
+        end
+        if @hash[:latent_heat_recovery]
+          openstudio_ideal_air.setLatentHeatRecoveryEffectiveness(@hash[:latent_heat_recovery])
+        else
+          openstudio_ideal_air.setLatentHeatRecoveryEffectiveness(@@schema[:definitions][:IdealAirSystem][:properties][:latent_heat_recovery][:default]) #TODO : Openstudio defaults are different from schema
+        end
+        if @hash[:demand_control_ventilation]
+          if @hash[:demand_control_ventilation] == true
+            openstudio_ideal_air.setDemandControlledVentilationType('OccupancySchedule') #TODO: when true, what is demand control vent. type?
```
tanushree04: @chriswmackey Is the demand_control_ventilation type `OccupancySchedule` when true?

lib/ladybug/energy_model/setpoint_humidistat_abridged.rb
chriswmackey: I'm not sure I understand your question.  Are you asking whether the schema allows the assignment of a People definition to a Room that isn't used in any of the ProgramTypes? If so the answer is yes, very much so.  The primary reason why the schema allows you to specify a custom object for a Room is to handle cases where just one or two Rooms are slightly different from the others. This way, you don't need to create a whole new ProgramType schema just for those Rooms.

lib/ladybug/energy_model/ideal_air_system.rb
chriswmackey: Yes. Here is the Python code:
https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/idealair.py#L286

chriswmackey:
Thanks @tanushree04 .  I responded quickly to your comments and I will do a longer review once I get the chance.

lib/ladybug/energy_model/space_type.rb
```diff
@@ -0,0 +1,127 @@
...
+      
+        if @hash[:ventilation]
+         ventilation = VentilationAbridged.new(@hash[:ventilation])
+         openstudio_ventilation = ventilation.to_openstudio(openstudio_model)
+         openstudio_space_type.setDesignSpecificationOutdoorAir(openstudio_ventilation)
+        end
+
+        if @hash[:setpoint]
+          setpoint_thermostat = SetpointThermostat.new(@hash[:setpoint])
+          openstudio_setpoint_thermostat = setpoint_thermostat.to_openstudio(openstudio_model)
```
chriswmackey: That is great that you are creating the thermostat but it looks like you never end up assigning a unique thermostat derived from this one to each thermal zone. So, ultimately, we just end up with a thermostat without a parent Zone in the Model, which E+ will throw an exception about.

Ultimately, I think that this code does not belong here because OpenStudio treats thermostats as objects that are unique to each Zone, which is fundamentally different from how the Honeybee schema allows you to create a Setpoint object that can be assigned to a Program Type.

So, rather than having this code here, I think you will need to add a method at the end of the model creation that loops through each Zone and sees if there's an ideal air system on the Zone but no thermostat. If so, you will have to find the hash of the ProgramType that is assigned to the Room, get the Setpoint hash, and use it to create a new unique thermostat for the Zone.

lib/ladybug/energy_model/space_type.rb
```diff
@@ -0,0 +1,127 @@
...
+        end
+
+        if @hash[:setpoint][:humidification_schedule] or @hash[:setpoint][:dehumidification_schedule]
+          setpoint_humidistat = SetpointHumidistat.new(@hash[:setpoint])
+          openstudio_setpoint_humidistat = setpoint_humidistat.to_openstudio(openstudio_model)
```
chriswmackey: Same comment about thermostats applies here to humidistats.

lib/ladybug/energy_model/ventilation_abridged.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+        if @hash[:flow_per_person]
+          openstudio_ventilation.setOutdoorAirFlowperPerson(@hash[:flow_per_person])
+        else 
+          openstudio_ventilation.setOutdoorAirFlowperPerson(@@schema[:definitions][:VentilationAbridged][:properties][:flow_per_person][:default])
+        end
+        if @hash[:flow_per_area]
+          openstudio_ventilation.setOutdoorAirFlowperFloorArea(@hash[:flow_per_area])
+        else 
+          openstudio_ventilation.setOutdoorAirFlowperPerson(@@schema[:definitions][:VentilationAbridged][:properties][:flow_per_area][:default])
```
chriswmackey: typo here

lib/ladybug/energy_model/ideal_air_system.rb
```diff
@@ -0,0 +1,99 @@
...
+      def defaults
+        result = {}
+        result
+      end
+
+      def create_openstudio_object(openstudio_model)       
+        openstudio_ideal_air = OpenStudio::Model::ZoneHVACIdealLoadsAirSystem.new(openstudio_model)
+        if @hash[:heating_limit]
+          openstudio_ideal_air.setHeatingLimit('LimitCapacity')
+          if @hash[:heating_limit] == 'autocalculate'
```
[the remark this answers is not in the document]
tanushree04: @chriswmackey in the python code https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/idealair.py#L254-L266 when no limit is specified, the limit type is `NoLimit` and the`MaximumSensibleHeatingCapacity` is blank.

When there is no `heating_limit` specified is the limit type still `LimitCapacity`, and the Maximum sensible heatign capacity should be autosized?

lib/ladybug/energy_model/model.rb
```diff
@@ -196,15 +204,77 @@ def create_global_construction_set
...
+        if @hash[:rooms]
+          $room_array_setpoint = [] 
           @hash[:rooms].each do |room|
           room_object = Room.new(room)
-          room_object.to_openstudio(@openstudio_model)
+          openstudio_room = room_object.to_openstudio(@openstudio_model)
+          if room[:properties][:energy][:program_type] && !room[:properties][:energy][:setpoint]
+            $room_array_setpoint << room
+            if $setpoint_array
+              $setpoint_array.each do |setpoint|
```
chriswmackey: I'm having difficulty understanding what this variable is exactly. Is is an array of all setpoint hashes on ProgramTypes throughout the model?  How does it differ from `room_array_setpoint`?  Why does it appear that you are looping through it inside the for loop that goes through each of the Room hashes? Or am I just not reading the Ruby syntax correctly here?
This is where some in-line comments would be really helpful for understanding what the different sections of your code is doing and what the variables are.

tanushree04: @chriswmackey its a global hash created here https://github.com/ladybug-tools-in2/energy-model-measure/blob/7cbaf0701a9a4a45829eef6efdb4186c6bf31ea1/lib/ladybug/energy_model/space_type.rb#L111

PR #12, opened 2019-11-20
feat(SimulationParameter): Added simulation parameter measure and tests

tanushree04:
@chriswmackey This PR includes methods to create OS SimulationParameter and tests to validate the example files against the simulation parameter schema. 

Some properties from the schema do not have methods to translate to OpenStudio objects. I have added those as comments. 

I will need to add the measure for translating the simulationparameter as well.


lib/ladybug/energy_model/simulation_parameter.rb
```diff
@@ -0,0 +1,188 @@
...
+          if @hash[:shadow_calculation][:calculation_frequency]
+            openstudio_shadow_calculation.setCalculationFrequency(@hash[:shadow_calculation][:calculation_frequency])
+          else
+            openstudio_shadow_calculation.setCalculationFrequency(@@schema[:definitions][:ShadowCalculation][:properties][:calculation_frequency][:default])
+          end
+          if @hash[:shadow_calculation][:maximum_figures]
+            openstudio_shadow_calculation.setMaximumFiguresInShadowOverlapCalculations(@hash[:shadow_calculation][:maximum_figures])
+          else
+            openstudio_shadow_calculation.setMaximumFiguresInShadowOverlapCalculations(@@schema[:definitions][:ShadowCalculation][:properties][:maximum_figures][:default])
+          end
```
tanushree04: To map the `calculation_method` from the schema, the `setCalculationMethod` in the OpenStudio SDK is supported by v. 2.9 and is not available for the current OpenStudio version.

lib/files/schema/simulationparameter.json
```diff
@@ -0,0 +1,313 @@
...
+          },
+          "include_sqlite": {
+            "title": "Include_Sqlite",
+            "default": true,
+            "type": "boolean"
+          },
+          "include_html": {
+            "title": "Include_Html",
+            "default": false,
+            "type": "boolean"
```
tanushree04: Currently there is no way not to include the html, and sqlite reports.

lib/files/schema/simulationparameter.json
```diff
@@ -0,0 +1,313 @@
...
+            }
+          },
+          "summary_reports": {
+            "title": "Summary_Reports",
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          }
+        },
```
tanushree04: Currently there are no methods to edit the summary reports in the OpenStudio SDK. There is an EnergyPlus measure to edit them.

lib/files/schema/simulationparameter.json
```diff
@@ -0,0 +1,313 @@
...
+              "Friday",
+              "Saturday"
+            ],
+            "type": "string"
+          },
+          "holidays": {
+            "title": "Holidays",
+            "type": "array",
+            "items": {
+              "$ref": "#/definitions/Date"
```
tanushree04: OpenStudio currently does not support holidays.

lib/ladybug/energy_model/simulation_parameter.rb
chriswmackey: I'm aware that this is a feature that Dan added before he left.  We are fine with only supporting 2.9 and above.

lib/files/schema/simulationparameter.json
chriswmackey: That's a bit of a shame because it forces the generation of potentially heavy files that might not be used by the simulator. This might be something we can live with, though.

lib/files/schema/simulationparameter.json
chriswmackey: If that is the case, we might need to ship the SummaryReport EnergyPlus measure you mention along with this energy-model-measure and have the energy-model-measure assign the E+ measure to the OpenStudio model whenever there are summary reports so that it gets executed upon the simulation.  Alternatively, if we have any remaining IN2 hours for Task 3, this would be a really helpful thing to have in the OpenStudio SDK.

lib/files/schema/simulationparameter.json
chriswmackey: I must have forgotten that we have holidays implemented in honeybee-legacy through a hack. Did we also not implement `apply_holiday` on the `ScheduleRuleset` schema, then? This really seems like a critical feature that the OpenStudio SDK should support since it is really not possible to simulate a lot of common building programs like religious centers or assembly halls without holidays.  I would say that this is probably the highest priority for any task 3 hours that remain and, if we don't have any hours left here, I guess we'll have to write an EnergyPlus measure to do this.

chriswmackey:
Thanks for putting this together @tanushree04 .  I will do a more detailed review shortly and I left some comments on the features that cannot be implemented in the meantime.

lib/files/schema/simulationparameter.json
tanushree04: @chriswmackey , I did not implement this in the measure for `ScheduleRuleset`, I do have `apply_holiday` in the schema. Joe had about 30 hours out of the 80 for task 3. He was using core openstudio development to charge his time as well. 
I can ask him how  much time it would take for him to holidays in the OpenStudio SDK.

lib/files/schema/simulationparameter.json
tanushree04: I can ask Joe about implementing this in the OS SDK as well.

lib/files/schema/simulationparameter.json
chriswmackey: That would be great if some of these features are address-able within the 30 hours.  I am also coming to realize that I'll likely only need 5 of the 50 hours that are budgeted to Matt for the standards gem help.  So it seems worthwhile to send an email to Kristin to see if we might be able to reallocate some of those original standards gem hours to addressing these issues.

tanushree04: I spoke to Joe, he has used up the hours on task 3. Matt and Marley collectively had 40 hours allocated. I am not sure aout Joe's availability but the remained of those could be reallocated. We can schedule another call to discuss this,

lib/ladybug/energy_model/simulation_parameter.rb
```diff
@@ -0,0 +1,188 @@
...
+       @@extension = ExtensionSimulationParameter.new
+       @@schema = nil
+       File.open(@@extension.schema_file) do |f|
+        @@schema = JSON.parse(f.read, symbolize_names: true)
+       end
+
+       @hash = hash
+       @type = @hash[:type]
+       raise 'Unknown model type' if @type.nil?
+       raise "Incorrect model type '#{@type}'" unless @type == 'SimulationParameter'
```
chriswmackey: I think these two lines should say something more generic like:
```
       raise 'Unknown object type' if @type.nil?
       raise "Incorrect type for SimulationParameter '#{@type}'" unless @type == 'SimulationParameter'

File: lib/ladybug/energy_model/simulation_parameter.rb (hunk 1 above)
Comment (chriswmackey) [reply lib/ladybug/energy_model/simulation_parameter.rb]: Ah, wait, I didn't realize that we hadn't yet put the `calculation_method` in. Let's leave this unresolved until we can do that.

File: lib/ladybug/energy_model/simulation_parameter.rb [hunk 6]
```diff
@@ -0,0 +1,188 @@
...
+            openstudio_shadow_calculation.setCalculationFrequency(@hash[:shadow_calculation][:calculation_frequency])
+          else
+            openstudio_shadow_calculation.setCalculationFrequency(@@schema[:definitions][:ShadowCalculation][:properties][:calculation_frequency][:default])
+          end
+          if @hash[:shadow_calculation][:maximum_figures]
+            openstudio_shadow_calculation.setMaximumFiguresInShadowOverlapCalculations(@hash[:shadow_calculation][:maximum_figures])
+          else
+            openstudio_shadow_calculation.setMaximumFiguresInShadowOverlapCalculations(@@schema[:definitions][:ShadowCalculation][:properties][:maximum_figures][:default])
+          end
+        end
```
chriswmackey: Also, it looks like yo aren't assigning the `solar_distribution` property, which is something we included in the `shadow_calculation` schema here but I know EnergyPlus/OpenStudio stores this property on the `Building`. So I think you will have to get all Buildings in the OpenStudio Model and assign the `solar_distribution` property to them.

lib/files/schema/simulationparameter.json
chriswmackey: What did Joe use the hours to address? I haven't had any contact with him and I actually didn't know that he was taking over Task 3 from Dan until you mentioned it here. The last work I remember happening on Task 3 was Dan adding in the `setCalculationMethod` to the `ShadowCalculation` object, though I didn't know if Dan had used up all of the Task 3 hours there.  A call to sort all of this out would be helpful.

lib/ladybug/energy_model/simulation_parameter.rb
tanushree04: @chriswmackey it is a method for the `SimulationControl` https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-1.8.0-doc/model/html/classopenstudio_1_1model_1_1_simulation_control.html. I can add it there.

lib/files/schema/simulationparameter.json
chriswmackey: FYI, @tanushree04 , Julien addressed this with an edit to the IDF ForwardTranslator:
https://github.com/NREL/OpenStudio/issues/3786#issuecomment-558318610

So we should be able to support this capability but it's not something that you need to worry about here in the energy-model-measure since the ForwardTranslator is not a part of the OpenStudio model object. Instead, it's something that we will address in the IDF translation process and I have some ideas for how to make use of this from Grasshopper if I find I can't do it through OpenStudio CLI.

Also, judging from the code that Julien pointed me towards, I think the exposure of SummaryReports is also going to be something that gets addressed in the ForwardTranslator object and not as a part of the energy-model-measure.

lib/ladybug/energy_model/simulation_parameter.rb
chriswmackey: Ah, I didn't know that's how it is in OpenStudio SDK.  Clearly, all of us had the same reaction that it was out of place on the EnergyPlus Building object. I see that a bunch of the other fields that are on EnergyPlus Buildings are on the OpenStudio SimulationControl.  Thanks for the link and this is good to know for the future.

tanushree04:
@chriswmackey The latest changes to scheduleruleset measure and example files are included in the last commit.

chriswmackey:
@tanushree04 ,

You know better than I which of these open comments you have addressed. Can you please resolve the ones that you have addressed with changes to the code and, if they are all done, we can merge this in.

tanushree04:
@chriswmackey the only unresolved comments now are about implementing holidays in SimulationParameter, and adding CalculationMethod in OpenStudio v2.9 onwards and editing SummaryReports.

chriswmackey:
Given that I need to make the Grasshopper component that uses this measure, I am just going to merge this in.  We can address the shadow Calculation in another PR.

PR #16, opened 2019-12-20
fix(shadow_calculation): Added calculation_method

tanushree04:


lib/ladybug/energy_model/model.rb
```diff
@@ -281,7 +281,7 @@ def create_rooms
               end
             end
             end
-          end
+          end
```
chriswmackey: You probably don't actually want to add this extra space here.

chriswmackey:
Also, your commit log is pretty long for such a small change here. We might want to squashand merge this one.

PR #17, opened 2019-12-21
fix(measure): Fix incorrect import paths throughout measure

chriswmackey:
This commit includes all of the changes that were needed to get the energy-model-measure to run with OpenStudio CLI. This included fixing all of the import paths that assumed ladybug was the root and fixing the xml file of the ladybug_simulation_parameter_measure. It also included commenting out the measure validation step since the json-schema gem is not shipped with OpenStudio CLI's Ruby environment.

After we merge this in, @tanushree04 , it would be great if you can make the validation step optional and ensure that it doesn't fail when there's no json-schema gem found.


lib/ladybug/energy_model/extension.rb
```diff
@@ -50,13 +50,13 @@ def initialize
 
       # Return the absolute path of the measures or nil if there is none, can be used when configuring OSWs
       def measures_dir
-        File.absolute_path(File.join(@root_dir, 'lib/measures/'))
+        File.absolute_path(File.join(@root_dir, 'OpenStudio/measures/'))
```
tanushree04: @chriswmackey in the GitHub Repo, the file path is lib/measures/

chriswmackey: Ok. I'll change this to be realtive instead of absolute since I think we will change the name from lib at some point. Do you have any idea why lib was chosen in the first place?

chriswmackey: Done. It's using a relative reference now.

lib/ladybug/energy_model.rb
```diff
@@ -29,41 +29,41 @@
...
-require 'ladybug/energy_model/electric_equipment_abridged'
-require 'ladybug/energy_model/gas_equipment_abridged'
-require 'ladybug/energy_model/infiltration_abridged'
-require 'ladybug/energy_model/ventilation_abridged'
-require 'ladybug/energy_model/setpoint_thermostat'
-require 'ladybug/energy_model/setpoint_humidistat'
-require 'ladybug/energy_model/ideal_air_system'
-require 'ladybug/energy_model/simulation_parameter'
\ No newline at end of file
+require "#{File.dirname(__FILE__)}/version"
```
tanushree04: @chriswmackey As per the folder structure, the address for this should be require `"#{File.dirname(__FILE__)}/energy_model/version"` .

PR #18, opened 2019-12-23
Revert "fix(shadow_calculation): Added calculation_method"

tanushree04:
Reverts ladybug-tools-in2/energy-model-measure#16


chriswmackey:
Hey @tanushree04 . Why was this PR reverted? The code looked good to me and I see that it fixed an issue that is currently preventing me from running the IDF through EnergyPlus.

PR #20, opened 2019-12-26
fix(conditional validation)- Added validation for schema if json-schema gem is present and loaded.

tanushree04:


lib/ladybug/energy_model/model_object.rb
```diff
@@ -29,8 +29,8 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *******************************************************************************
 
-require 'json'
 require 'openstudio'
+require 'json-schema'
```
chriswmackey: This will cause a failure

PR #30, opened 2019-12-28
fix(ladybug): Fix paths to use the ladybug folder as a gem

chriswmackey:
It seems that the repo here has actually been structured as a Ruby gem, contrary to what we thought in our last meeting.  All I had to do to make it a valid gem was add a `ladybu.rb` file and now I am able to pass in ladybug as a gem to the OpenStudio CLI.

This commit also includes a fix to the previous PR where we were making the schema validation optional.  Now, there's an extra check is now performed to see if the `json-schema` gem is available in the `valid?` call.

Lastly, I added some extra comments now that I understand that @macumber structured this repo as an OpenStudio extension.


lib/ladybug/energy_model/extension.rb
```diff
@@ -83,7 +90,7 @@ def schema
         @@schema
       end
 
-    # check if the schema is valid
+    # check if the model schema is valid
       def schema_valid?
         if Gem.loaded_specs.has_key?("json-schema")
           require 'json-schema'
```
tanushree04: @chriswmackey should there be an `else` statement here as well?

chriswmackey: I think if this function returned something, we would need the `else` statement.  But, because it's just running the OpenAPI schema through the JSON validator only when the `json-schema` is present (and not doing anything when it's not there), it should be fine.

chriswmackey:
Great! Thank you @tanushree04 .

PR #60, opened 2020-02-06
fix(face): Fix bug in previous PR

chriswmackey:
I'm guessing that the tests weren't re-run after the last PR.  This is why we really need to get continuous integration set up since I know it's not realistic to trust ourselves that we'll always run the tests locally.


tanushree04:
@chriswmackey I resolved the merge conflicts using GitHub and running the tests locally didn't catch the errors after merging. The CI would help with this process.

chriswmackey:
I see. That explains why we missed the bug if you resolved conflicts within the browser. Continuous integration would catch the error in this case.

PR #61, opened 2020-02-07
fix(tests): Fix incorrect tests and document how to run them

chriswmackey:
This commit makes a number of changes to the test-running process including:

* Get rid of tests checking the re-serialization of the same file to a Model twice. These tests were pretty clearly for the older to_openstudio method, which had functionality we were not using and we have removed.
* Get rid of sample JSONs using outdated versions of the schema.
* Updated the version of the OpenAPI specification used by the measure
* Added documentation to the README.MD explaining how to install dependencies and run the tests
* Ensured we use the name of honeybee schema objects instead of openstudio ones (eg. program type and not space type)

Once we merge this in, I will start trying to implement continuous integration to automatically run the tests every time that we send a PR.

Also I think I'll restructure the [samples of the honeybee-schema repo](https://github.com/ladybug-tools/honeybee-schema/tree/master/samples) to have the same folder organization as we have here for the tests.


tanushree04:
@chriswmackey This looks good to merge to me.

chriswmackey:
Thank you @tanushree04 . I'll merge it and then I will see if I can get continuous integration set up today.

Issue #72, opened 2020-02-11
Add some assertion checks to the tests


chriswmackey:
Most of our tests right now consist of parsing in the sample JSONs just to make sure that they pass.  I know from [comments like this one](https://github.com/ladybug-tools-in2/energy-model-measure/blob/master/spec/tests/honeybee_hvac_spec.rb#L56) that we also have plans to implement some assertion checks to verify that the properties of the OpenStudio objects have been set to the values specified in the JSON (and that the defaults have been set correctly).   Adding checks like this on larger objects (like Models) will likely be unnecessarily time consuming but we should at least add some checks like this for smaller objects (materials, schedules, ideal air systems). I'm sure that the process is likely to bring some typos/bugs to the surface if my past experience is any indication.

chriswmackey:
I think we can close this out for now.

PR #74, opened 2020-02-18
Addresses issues in measure. 

### LINKED ISSUE #13: Implement translation of holidays from honeybee RunPeriod to OpenStudio once the OS SDK supports it
At the time that I am writing this issue, the OpenStudio SDK does not support applying holidays to a simulation.  This should hopefully be addressed in the next OpenStudio release and, once it is addressed, this repo should be updated.

tanushree04:
Closes 
#13 : Includes methods to implement holiday_schedule, and updates schema with schedule_holiday. 
#72 : Includes assertion tests to check whether properties of OpenStudio objects are correctly assigned.

### DIFF
--- a/lib/from_honeybee/geometry/aperture.rb
+++ b/lib/from_honeybee/geometry/aperture.rb
@@ -60,7 +60,15 @@ module FromHoneybee
       end
       reordered_vertices = OpenStudio.reorderULC(os_vertices)
 
+<<<<<<< HEAD
       os_subsurface = OpenStudio::Model::SubSurface.new(reordered_vertices, openstudio_model)
+=======
+      os_subsurface = OpenStudio::Model::SubSurface.new(os_vertices, openstudio_model)
+      os_vertices = os_subsurface.vertices
+      # reorder vertices
+      reordered_vertices = OpenStudio.reorderULC(os_vertices)
+      os_subsurface.setVertices(reordered_vertices)
+>>>>>>> Implemented reorder vertices for aperture, door, shade
       os_subsurface.setName(@hash[:name])
 
       # assign the construction if it exists
--- a/lib/from_honeybee/geometry/aperture.rb
+++ b/lib/from_honeybee/geometry/aperture.rb
@@ -60,11 +60,7 @@ module FromHoneybee
       end
       reordered_vertices = OpenStudio.reorderULC(os_vertices)
 
-      os_subsurface = OpenStudio::Model::SubSurface.new(os_vertices, openstudio_model)
-      os_vertices = os_subsurface.vertices
-      # reorder vertices
-      reordered_vertices = OpenStudio.reorderULC(os_vertices)
-      os_subsurface.setVertices(reordered_vertices)
+      os_subsurface = OpenStudio::Model::SubSurface.new(reordered_vertices, openstudio_model)
       os_subsurface.setName(@hash[:name])
 
       # assign the construction if it exists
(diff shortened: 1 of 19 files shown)


chriswmackey:
Thanks for doing this @tanushree04 . We should get the tests to pass before we merge.  I can see that [it's failing](https://travis-ci.org/ladybug-tools-in2/energy-model-measure/builds/652235430?utm_source=github_status&utm_medium=notification) seemingly because of all the changes that @nllong has been pushing to the OpenStudio extension gem:
https://github.com/NREL/openstudio-extension-gem/commits/develop

@DavidGoldwasser , do you know that reason for this? Is this rush of changes to the extension gem going to be finished at some point? And, I assume we'll need to upgrade the version of Ruby and OpenStudio we are using once it's done.

lib/from_honeybee/geometry/aperture.rb
```diff
@@ -60,7 +60,11 @@ def to_openstudio(openstudio_model)
       end
       reordered_vertices = OpenStudio.reorderULC(os_vertices)
 
-      os_subsurface = OpenStudio::Model::SubSurface.new(reordered_vertices, openstudio_model)
+      os_subsurface = OpenStudio::Model::SubSurface.new(os_vertices, openstudio_model)
+      os_vertices = os_subsurface.vertices
+      # reorder vertices
+      reordered_vertices = OpenStudio.reorderULC(os_vertices)
+      os_subsurface.setVertices(reordered_vertices)
```
chriswmackey: Why did you make this change?  It looks like you are now running the reorder command twice and you have turned a single, efficient line of code into 5. Was there some reason that the previous code was not working? If not, can you change it back for all 4 cases that this code appears?

lib/from_honeybee/schedule/ruleset.rb
```diff
@@ -71,6 +71,15 @@ def to_openstudio(openstudio_model)
...
       end
+
+      # assign holiday schedule
+      if @hash[:holiday_schedule]
+        holiday_schedule = openstudio_model.getScheduleDayByName(@hash[:holiday_schedule])
+        unless holiday_schedule.empty?
+          holiday_schedule_object = holiday_schedule.get
+          os_sch_ruleset.setHolidaySchedule(holiday_schedule_object)
+        end
+      end
```
chriswmackey: Can this be put in a try/except so that it can still work in older OpenStudio versions?

tanushree04: @chriswmackey I added a begin rescue statement that should run if the method exists.

lib/from_honeybee/geometry/face.rb
```diff
@@ -63,10 +63,9 @@ def to_openstudio(openstudio_model)
       end
       reordered_vertices = OpenStudio.reorderULC(os_vertices)
 
-      os_surface = OpenStudio::Model::Surface.new(reordered_vertices, openstudio_model)        
+      os_surface = OpenStudio::Model::Surface.new(reordered_vertices, openstudio_model)
```
tanushree04: @chriswmackey I am not sure why it changed to the earlier reordered_vertices statements. Maybe when I rebased my fork .

lib/from_honeybee/schedule/ruleset.rb
tanushree04: @chriswmackey The else statement was not necessart in this case.

chriswmackey:
Thanks @tanushree04 . Good to know. I'll update the travis file before the end of today and, if all of the tests pass, I'll merge this in.

PR #78, opened 2020-03-12
Fixes #77

### LINKED ISSUE #77: Ensure that Doors are Correctly Translated to OpenStudio Doors and not FixedWindow
We recently realized that opaque doors are being translated to FixedWindows by the energy-model-measure:
image

It's possible that this confusion results from the fact that EnergyPlus calls both windows and doors "Fenestration Surfaces".  But this should hopefully be a quick fix in the code.

@tanushree04 , Is this something that you might be able to address quickly?  It would be a huge help if so and that would allow me to focus efforts on the last things needed for dragonfly-URBANopt integration.

tanushree04:
Fixes bug while creating Doors. 
The `setSurface` property on the Door Subsurface was applied after the SurfaceType was set to "door" and it reset the SurfaceType to "Fixed Window" by default. 

The SurfaceType is specified after the setSurface now.

### DIFF
--- a/lib/from_honeybee/geometry/face.rb
+++ b/lib/from_honeybee/geometry/face.rb
@@ -128,14 +128,16 @@ module FromHoneybee
           honeybee_door = Door.new(door)
           os_subsurface_door = honeybee_door.to_openstudio(openstudio_model)
           os_subsurface_door.setSurface(os_surface)
-          if door[:is_glass] == false
-            # set subsurface type
-            os_subsurface_door.setSubSurfaceType('Door')
-          elsif door[:is_glass] == true
+          if door[:is_glass] == true
             os_subsurface_door.setSubSurfaceType('GlassDoor')
-          elsif @hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor' && @hash[:boundary_condition][:type] == 'Outdoors'
+          elsif (@hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor') && @hash[:boundary_condition][:type] == 'Outdoors'
             os_subsurface_door.setSubSurfaceType('OverheadDoor')
+          elsif door[:is_glass] == false or door[:is_glass].nil?
+            os_subsurface_door.setSubSurfaceType('Door')
+            puts "3HELLO"
           end
+          puts "1HELLO = #{door}"
+          puts "2HELLO = #{os_subsurface_door}"
         end
       end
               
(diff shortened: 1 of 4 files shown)


lib/from_honeybee/geometry/face.rb
```diff
@@ -125,12 +125,17 @@ def to_openstudio(openstudio_model)
...
-          os_subsurface_door = door.to_openstudio(openstudio_model)
-          if @hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor' && @hash[:boundary_condition][:type] == 'Outdoors'
-            os_subsurface_aperture.setSubSurfaceType('OverheadDoor')
-          end
+          honeybee_door = Door.new(door)
+          os_subsurface_door = honeybee_door.to_openstudio(openstudio_model)
           os_subsurface_door.setSurface(os_surface)
+          if door[:is_glass] == false
+            # set subsurface type
+            os_subsurface_door.setSubSurfaceType('Door')
```
tanushree04: Subsurface type is set after setting surface

tanushree04: subsurface type is set after setting the surface.

lib/from_honeybee/geometry/face.rb
```diff
@@ -125,12 +125,17 @@ def to_openstudio(openstudio_model)
...
+          elsif door[:is_glass] == true
+            os_subsurface_door.setSubSurfaceType('GlassDoor')
+          elsif @hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor' && @hash[:boundary_condition][:type] == 'Outdoors'
+            os_subsurface_door.setSubSurfaceType('OverheadDoor')
```
chriswmackey: This check needs to happen inside the `if door[:is_glass] == false` check. As it is written now, no door will ever get to this check and be assigned as an overhead door because it always has an `is_glass` property that is true or false. Or, if the door has no `is_glass` property, we should assume the default, which is an opaque door with `false` `is_glass`.

Also, I think you need to put parentheses around `(@hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor')`. As it's written now, it looks like an indoor `RoofCeiling` would still result in an overhead door.

tanushree04: @chriswmackey I added the parentheses for `(@hash[:face_type] == 'RoofCeiling' or @hash[:face_type] == 'Floor')`. 
I added `elsif door[:is_glass] == false or door[:is_glass].nil?` after setting `OverheadDoor` to set as `Door` only if it is not `OverheadDoor` and also set `OverheadDoor` correctly even if is_glass property is nil.

PR #83, opened 2020-03-20
Add summary reports to measure

tanushree04:


chriswmackey:
I don't see any changes related to summary reports in the "Files changed". As far as I can tell, I only see changes that we have already merged in, which means that you did not re-sync your fork with the master before adding your changes and sending the PR. Can you please re-sync your fork?

tanushree04:
@chriswmackey I added methods to create SummaryReports now. I also found some typos and made some other edits to the code.

tanushree04:
@chriswmackey The latest commit adds SummaryReports. I synced my fork with ladybug tools- master. I found some typos and made some other edits which you can see in the files changed for the earlier commit.

chriswmackey:
Thanks @tanushree04 . You will see that the tests are failing:
https://travis-ci.com/github/ladybug-tools-in2/energy-model-measure/builds/154635174

If you can fix those, we should be able to merge.

tanushree04:
@chriswmackey is this because the measure is running OpenStudio 2.9.1, and these methods are in 3.0?

chriswmackey:
@tanushree04 ,
Can we put this in a try/catch loop then? We'll still need the measure to work with version 2.9 for now and we can get rid of the try/catch once we officially support version 3.

chriswmackey:
Also, why are you deleting the 2 simulation parameter JSONs with this PR?

tanushree04:
@chriswmackey I can add the begin rescue loop.
The simulation parameter files are deleted because I updated the path in the measure test so that they are loaded from the [spec/samples folder](https://github.com/ladybug-tools-in2/energy-model-measure/blob/19a56d6a973bc02877ea2dbeff62e5749850223a/lib/measures/from_honeybee_simulation_parameter/tests/from_honeybee_simulation_parameter_test.rb#L103)

lib/from_honeybee/simulation/parameter.rb
```diff
@@ -200,6 +200,18 @@ def create_openstudio_objects
             end
           end
         end
+        if @hash[:output][:summary_reports]
+          @hash[:output][:summary_reports].each do |report|
+            begin
+              os_report = OpenStudio::Model::OutputTableSummaryReports.new(@openstudio_model)
+            rescue NameError
```
chriswmackey: Shouldn't this be outside of the "for" loop? It seems like the SDK is set up to just create one `OutputTableSummaryReports` object  to which you can add as many summary reports as you want.

tanushree04: Yes, I was interpreting it incorrectly

chriswmackey:
Thanks, @tanushree04 . Let's wait until the call to merge this. I wanted to show how PRs should be merged in order to correctly trigger the automated versioning.

PR #96, opened 2020-04-26
Initial implementation of sub surface triangulation

### LINKED ISSUE #58: Get measure to triangulate all Apertures and Doors with more than 4 vertices
I'm opening this issue so that we track it separately from the other geometry issues that we should address with the measure.  To paste in Dan's advice for triangulation with the OpenStudio SDK:

There is also a triangulation method:
https://github.com/NREL/OpenStudio/blob/2c4dd5c9d790d1e1a019b328ca06957660733478/src/utilities/geometry/Geometry.hpp#L99

This repo is an overview of how to do geometry stuff in OpenStudio:
https://github.com/macumber/coordinate-system-tutorial

I don't think we ever really cared to make "nice" triangulation of the holes/windows themselves, just triangulated them if they were > 4 vertices.  Typically we would just tag all the triangles with some id so we can associate them back to the original window. 

We can add more functionality to the API based on methods in boost geometry or polypartition.

macumber:
Here is an initial cut at the triangulation, let me know if you see any issues. Fixes #58

image


lib/from_honeybee/geometry/aperture.rb
```diff
@@ -59,39 +59,80 @@ def to_openstudio(openstudio_model)
...
+        tInv = t.inverse
+        face_vertices = OpenStudio::reverse(tInv*reordered_vertices)
+        
+        # no holes in the subsurface
+        holes = OpenStudio::Point3dVectorVector.new
+        
+        # triangulate surface
+        triangles = OpenStudio::computeTriangulation(face_vertices, holes)
+        if triangles.empty?
+          raise "Failed to triangulate aperture #{@hash[:identifier]} with #{reordered_vertices.size} vertices"
```
macumber, line 79: @chriswmackey the biggest question i had was how you want to handle errors in triangulation?

lib/from_honeybee/geometry/aperture.rb
```diff
@@ -59,39 +59,80 @@ def to_openstudio(openstudio_model)
...
-        adj_srf_identifier = @hash[:boundary_condition][:boundary_condition_objects][0]
-        sub_srf_ref = openstudio_model.getSubSurfaceByName(adj_srf_identifier)
-        unless sub_srf_ref.empty?
-          sub_srf = sub_srf_ref.get
-          os_subsurface.setAdjacentSubSurface(sub_srf)
+      # triangulate subsurface if neccesary
+      triangulated = false
+      final_vertices_list = []
+      if reordered_vertices.size > 4
+
```
chriswmackey, line 67: Can we add in a check that, if the geometry has a `Surface` boundary condition and needs to be triangulated, we raise an exception for now? Since you are not currently mapping the multiple triangles to the new adjacent triangles, the resulting model for this case right now would not be valid/simulate-able. In Python, we have methods that get the adjacent surface and triangulate the neighbor simultaneously and don't have this issue. Until I can implement those methods here, I think it's safer to raise an exception.

lib/from_honeybee/geometry/aperture.rb
chriswmackey: Is there a list of known cases that fail here?  I understand that there are always possible situations of poor input (like non-planar shapes or self-intersecting bow-tie shapes). But if it's anything like the ear clipping method that we currently use in Python, this shouldn't be failing for any of the cases that we permit from the plugin side. So I think raising an exception like this is fine.

lib/from_honeybee/geometry/aperture.rb
```diff
@@ -59,39 +59,80 @@ def to_openstudio(openstudio_model)
...
+        if @hash[:properties][:energy][:construction]
+          construction_identifier = @hash[:properties][:energy][:construction]
+          construction = openstudio_model.getConstructionByName(construction_identifier)
+          unless construction.empty?
+            os_construction = construction.get
+            os_subsurface.setConstruction(os_construction)
+          end
+        end
+        
+        # assign the bondary condition object if it's a Surface
```
chriswmackey, line 114: Typo: `boundary`

lib/from_honeybee/geometry/aperture.rb
```diff
@@ -59,39 +59,80 @@ def to_openstudio(openstudio_model)
...
+        # create new list of surfaces
+        triangles.each do |vertices|
+          final_vertices_list << OpenStudio::reverse(t*vertices)
         end
+        
+        triangulated = true
+        
+      else
+        # reordered_vertices are good as is
+        final_vertices_list << reordered_vertices
```
chriswmackey, line 91: Just to check here, are these vertices ordered by ULC?

lib/from_honeybee/geometry/aperture.rb
macumber: I could take a crack at implementing that, basically I would triangulate the first side, then copy reversed portions of the first window into the second zone (rather than re-triangulating the second sub surface).

macumber: Probably want a unit test or two for this case.

lib/from_honeybee/geometry/aperture.rb
macumber: This triangulation should be pretty robust but there is always some crazy edge case that will come up.

lib/from_honeybee/geometry/aperture.rb
macumber: Can you make a JSON similar to `model_5vertex_sub_faces.json` but with the aperture and door on interior surfaces (e.g. that require surface matching)?

lib/from_honeybee/geometry/aperture.rb
macumber: Yes, already reordered

lib/from_honeybee/geometry/aperture.rb
```diff
@@ -59,39 +59,110 @@ def to_openstudio(openstudio_model)
...
+      # triangulate subsurface if neccesary
+      triangulated = false
+      final_vertices_list = []
+      matching_os_subsurfaces = []
+      matching_os_subsurface_indices = []
+      if reordered_vertices.size > 4
+      
+        # if this apeture has a matched apeture, see if the other one has already been created
+        # the matched apeture should have been converted to multiple subsurfaces
+        if @hash[:boundary_condition][:type] == 'Surface'
```
macumber, line 72: @chriswmackey I think this new code will handle the case of adjacent interior subsurfaces that require triangulation.  That said it is completely untested.  It would be great if you could make an example JSON to unit test this case.

lib/from_honeybee/geometry/aperture.rb
chriswmackey: Understood. I think this is fine, then.

chriswmackey:
Thank you for doing this @macumber . It is really helpful to have the Surface boundary condition case taken care of. I will add a unit test now for interior sub-faces that need triangulation.

macumber:
@chriswmackey this is ready to go now!

lib/from_honeybee/geometry/aperture.rb
macumber: This is now tested, thanks for the test model!

macumber:
Note that both the OS App and SketchUp Plug-in have trouble rendering the resulting OSM.  However the geometry is correct and it does run through E+ correctly:

  | Construction | Area of One Opening [m2] | Area of Openings [m2] | Glass U-Factor [W/m2-K] | Glass SHGC | Glass Visible Transmittance | Parent Surface
-- | -- | -- | -- | -- | -- | -- | --
FRONTAPERTURE..0 | GENERIC SINGLE PANE | 1.50 | 1.50 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE1_FRONT
FRONTAPERTURE..1 | GENERIC SINGLE PANE | 1.50 | 1.50 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE1_FRONT
FRONTAPERTURE..2 | GENERIC SINGLE PANE | 0.40 | 0.40 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE1_FRONT
BACKAPERTURE..0 | GENERIC SINGLE PANE | 1.50 | 1.50 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE2_BACK
BACKAPERTURE..1 | GENERIC SINGLE PANE | 1.50 | 1.50 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE2_BACK
BACKAPERTURE..2 | GENERIC SINGLE PANE | 0.40 | 0.40 | 5.801 | 0.816 | 0.880 | TINYHOUSEZONE2_BACK
Total or Average |   |   | 6.80 | 5.801 | 0.816 | 0.880 |  

image

image

Issue #91, opened 2020-04-22
Add our own URBANopt Baseline.rb mapper to this repo


chriswmackey:
After I got the chance to talk to Dan Macumber and he explained the original idea for how we would integrate with URBANopt, I am realizing that we should refactor the way that we run this honeybee measure in URBANopt. There are several steps that would be much more streamlined if we just wrote out our own base_workflow.osw and corresponding Baseline.rb mapper rather than using the one that ships with URBANopt by default. For example:

* We could just write out the honeybee JSONs from the interface end and not have to worry about the translation to osm as a separate step. URBANopt SDK could take care of the translation as long as we included the [energy-model-measure](https://github.com/ladybug-tools-in2/energy-model-measure) in the base_workflow.osw.
* We could give the user the ability to customize the measures that went into the base_workflow.osw on the interface end. This would allow them the option of either hard-assigning measure inputs as you would for a single OpenStudio model or use one of the keys in the geoJSON feature file as the input if they so desired. We would just add the users's input to the  base_workflow.osw that is auto-generated by our core libraries and all would be good.

To enable, this, I think we would have to add a Baseline.rb mapper to the energy_model_measure repo. I am thinking that we can add it to the `files` folder so that it still ships with the Ruby gem but it's not a critical element of the `from_honeybee` library.

@tanushree04 ,
Are there any resources that you could point me towards for writing one's own URBANopt mapper? I could just start by deleting stuff out from the Baseline.rb mapper that ships with URBANopt if not.

tanushree04:
@chriswmackey as mentioned in the call, the [Gemfile](https://github.com/urbanopt/urbanopt-example-geojson-project/blob/develop/Gemfile) dictates the gems which gems are installed in the .bundle. 
The URBANopt documentation https://docs.urbanopt.net/customization/modifying_mapper.html has some information on the functionality of the Mapper classes. The best way would be to use the existing mapper class and removing what is not necessary, and add the mapping of the honeybee JSON to the energy model measure argument.

chriswmackey:
It would be very helpful if you can take the first pass at this @tanushree04 .  Thank you, as always.

chriswmackey:
Closing this now that this has been addressed.

PR #97, opened 2020-04-27
feat(mapper class): Add mapper class and osw file

tanushree04:
Addresses #91

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ else
   gem 'openstudio-extension', github: 'NREL/OpenStudio-extension-gem', tag: 'v0.1.4'
 end
 
-# this code appears to be outdated and so it's commented out for now
+# this code appears to be outdated and so it's commented out for now. Uncomment to run tests locally
+#using `bundle exec rake`.
 # simplecov has an unneccesary dependency on native json gem, use fork that does not require this
-gem 'simplecov', github: 'NREL/simplecov'
+#gem 'simplecov', github: 'NREL/simplecov'
--- a/Gemfile
+++ b/Gemfile
@@ -11,4 +11,4 @@ end
 
 # this code appears to be outdated and so it's commented out for now
 # simplecov has an unneccesary dependency on native json gem, use fork that does not require this
-# gem 'simplecov', github: 'NREL/simplecov'
+gem 'simplecov', github: 'NREL/simplecov'
--- a/lib/from_honeybee/_openapi/HoneybeeMapper.rb
+++ b/lib/from_honeybee/_openapi/HoneybeeMapper.rb
@@ -1,0 +1,127 @@
+# *******************************************************************************
+# Honeybee Energy Model Measure, Copyright (c) 2020, Alliance for Sustainable 
+# Energy, LLC, Ladybug Tools LLC and other contributors. All rights reserved.
+# 
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# (1) Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# (2) Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# (3) Neither the name of the copyright holder nor the names of any contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission from the respective party.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER(S) AND ANY CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+# THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER(S), ANY CONTRIBUTORS, THE
+# UNITED STATES GOVERNMENT, OR THE UNITED STATES DEPARTMENT OF ENERGY, NOR ANY OF
+# THEIR EMPLOYEES, BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *******************************************************************************
+
+require 'urbanopt/scenario'
+require 'openstudio/common_measures'
+require 'openstudio/model_articulation'
+
+require 'json'
+
+module URBANopt
+    module Scenario
+      class HoneybeeMapper < SimulationMapperBase
+
+      # class level variables
+      @@instance_lock = Mutex.new
+      @@osw = nil
+      @@geometry = nil
+
+      def initialize()
+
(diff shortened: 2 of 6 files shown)


lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+            # required before this
+            @@osw = OpenStudio::Extension.configure_osw(@@osw)
+            end
+          end
+      end
+
+      def create_osw(scenario, features, feature_names)
+
+        if features.size != 1
+          raise "TestMapper1 currently cannot simulate more than one feature"
```
chriswmackey, line 69: What is the reason for this limitation? Is the only thing that we need to do to get rid of it is put all the code below this into a `for each` loop?

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        # deep clone of @@osw before we configure it
+        osw = Marshal.load(Marshal.dump(@@osw))
+
+        # now we have the feature, we can look up its properties and set arguments in the OSW
+        osw[:name] = feature_name
+        osw[:description] = feature_name
+        
+        # point to measures directory in energy-model-measure
+        osw[:measure_paths] << File.join(File.dirname(__FILE__) + "/../../measures")
```
chriswmackey, line 96: I am happy to see that there's an attempt being made to add the measure paths for the energy_model_measure but I think something hard-coded like this is not going to work for most URBANopt project folder cases if we have to copy this mapper into the project folder.

Alternatively, is there some way to call the URBANopt CLI and pass it this mapper without having to copy it into the project folder? If so, can you add documentation about how to do this to the ReadMe of this repo?  If not, can you just remove this line and we will take care of setting the measure paths when we write out the base_workflow.osw from our Python libraries.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        # store name of input JSON file for feature
+        feature_json = feature.detailed_model_filename
+
+        #TODO: Update path
+        simulation_parameter_path = File.join(File.dirname(__FILE__) + "/../../../spec/samples/model/#{feature_json}")
```
chriswmackey, line 102: I don't think we want to map different simulation parameters to different features in the geoJSON. There should really only be one of these simulation parameters for all of the features. Can you remove this line?

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        #TODO: Update path
+        model_path = File.join(File.dirname(__FILE__) + "/../../../spec/samples/simulation_parameter/#{feature_json}")
```
chriswmackey, line 105: Why isn't this just set to be equal to the `feature_json`?

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        if feature_type == 'Building'
+
+            OpenStudio::Extension.set_measure_argument(osw, 'from_honeybee_simulation_parameter', 'simulation_parameter_json', simulation_parameter_path)
```
chriswmackey, line 109: As I said, I think we can remove this line and we will just write out one set of simulation parameters for all of the features.

lib/from_honeybee/_openapi/honeybee_workflow.osw
```diff
@@ -0,0 +1,47 @@
...
+  "seed_file": null,
+  "weather_file": null,
+  "measure_paths": [],
+  "file_paths": [],
+  "run_directory": null,
+  "steps": [
+    {
+      "measure_dir_name": "from_honeybee_simulation_parameter",
+      "arguments": {
+        "simulation_parameter_json": "multi_zone_model.json"
```
chriswmackey, line 11: I know this osw is just as an example since we will write our own out from the Python end but can we change this line to something more meaningful like:
```
"simulation_parameter_json": "simulation_parameter.json"
```

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
...
+        if features.size != 1
+          raise "TestMapper1 currently cannot simulate more than one feature"
+        end
+        feature = features[0]
+        feature_id = feature.id
+        feature_type = feature.type
+        feature_name = feature.name
+        if feature_names.size == 1
+          feature_name = feature_names[0]
+        end
```
chriswmackey, line 77: What is the purpose of this `if` statement here? Can we remove it?

lib/from_honeybee/_openapi/honeybee_workflow.osw
```diff
@@ -0,0 +1,47 @@
...
+        "add_swh": false,
+        "add_refrigeration": false,
+        "remove_objects": false,
+        "system_type": "Inferred",
+        "add_hvac": true,
+        "use_upstream_args": false
+      }
+    },{
+      "measure_dir_name": "default_feature_reports",
+      "arguments": {
```
chriswmackey, line 38: I realize that this osw is mostly for demonstration purposes but I think we will `__SKIP__` this measure by default and only run it as a separate step when it is requested.

Gemfile
```diff
@@ -11,4 +11,4 @@ end
 
 # this code appears to be outdated and so it's commented out for now
 # simplecov has an unneccesary dependency on native json gem, use fork that does not require this
-# gem 'simplecov', github: 'NREL/simplecov'
+gem 'simplecov', github: 'NREL/simplecov'
```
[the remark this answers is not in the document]
chriswmackey: Or maybe just add a comment for where this dependency issue is relevant? Everything seems to be working fine when it's commented out.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
tanushree04: Currently the Mapper is creating an OSW file for each feature. The `create_osw` method is passing an array which consists of only 1 feature and is therefore getting the first element from the array.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
tanushree04: You could add the path of the feature_json in the detailed_model_filename property for each feature, or store the model_name.json in the detailed_model_filename and add the common path for each feature_json in the mapper class. I can make this equal to feature_json for now.

tanushree04: You could add the path of the feature_json in the detailed_model_filename property for each feature, or store the model_name.json in the detailed_model_filename and add the common path for each feature_json in the mapper class. I can make this equal to feature_json for now.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
chriswmackey: Ah, I think I see now. So the ability to pass multiple features is just for some future workflow where you might want to generate a single osw for multiple features (like if all of the buildings need to go into a single OpenStudio model). Ok. This is good, then.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
chriswmackey: I see your point but I think we should just make this equal to the `feature_json` until we have some specific folder structure that we want to use.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
chriswmackey: I see that the same code is in the baseline mapper so it's fine if you want to leave it in when we merge. It might be good to know why it's there if you find out, though.

chriswmackey:
I responded to the unresolved issues, @tanushree04 .  It looks like it should just be 2 quick tweaks and then we cna merge.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
```diff
@@ -0,0 +1,127 @@
+# *******************************************************************************
+# Honeybee Energy Model Measure, Copyright (c) 2020, Alliance for Sustainable
```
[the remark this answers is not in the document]
chriswmackey: It looks like this one has not been addressed yet so I unresolved it.

lib/from_honeybee/_openapi/HoneybeeMapper.rb
tanushree04: This was added for a similar reason, to possible extend create_osw to take more than 1 feature. Currently it is passed in as an array but is taking just the first feature_name from it.

PR #113, opened 2020-05-27
ci(semantic-release): replace gemspec version on release

### LINKED ISSUE #112: Set up semantic-release deploy to ruby gems
@chriswmackey  asked me to check out whether we could set up continuous deployment to RubyGems versioned by the semantic release script.

AntoineDao:
The deploy job has a script that replaces the gemspec version by the new tag ($TRAVIS_TAG)

close #112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,13 @@ jobs:
     - git config --global user.email "releases@ladybug.tools"
     - git config --global user.name "ladybugbot"
     - npx semantic-release
-  - stage: docs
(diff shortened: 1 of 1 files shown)


chriswmackey:
@AntoineDao ,
You are wonderful!  Thank you for figuring this out.  It probably would have taken me at least half a day to do this.

I am going to merge this and then I'll replace the hard-coded version number in the library with something that distinguishes it from the rubygems version number that you are setting here.

Then, I will upgrade all of our dependencies to their latest version.

Issue #98, opened 2020-04-29
Failed to run with the new OpenStudio 3.0


MingboPeng:
After updated to the newest OpenStudio 3.0, I got this error:

![image](https://user-images.githubusercontent.com/9031066/80647355-52cc1a00-8a3c-11ea-88c9-627b126b65e6.png)

chriswmackey:
It looks like the whole set of SDK methods changed for this object in OpenStudio 3.0:
https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.0.0-doc/model/html/classopenstudio_1_1model_1_1_shadow_calculation.html

It looks like the method is now called `setShadingCalculationUpdateFrequencyMethod`

MingboPeng:
@chriswmackey is this going to be fixed soon? or we are not going to support OpenStudio 3.0 yet?

Issue #15, opened 2019-12-05
Test the measure using OpenStudio 3.0 and the new version of Ruby it uses


chriswmackey:
In theory, everything should work but we should nevertheless test it.

chriswmackey:
I started doing some tests since @tanushree04 pointed me towards the 3.0 beta. The first issue that has popped up is that the `setCalculationFrequency` method seems to have changed in 3.0 since I get the following error when I run the measure:
```
[16:22:18.941581 ERROR] undefined method `setCalculationFrequency' for #<OpenStudio::Model::ShadowCalculation:0x00000241f32e7840>
```
@tanushree04 , do you know what this has changed to in version 3.0? Or is this a bug in the pre release?

tanushree04:
@chriswmackey I think the method is renamed https://github.com/NREL/OpenStudio/blob/e42fde8aa21cf78bc8d961dcbc8480448f50e102/src/model/ShadowCalculation.hpp#L115

chriswmackey:
Ok. We'll just have to keep a running log of changes here until we are ready to update the measure completely to 3.0

PR #116, opened 2020-05-27
feat(openstudio): Upgrade to OpenStudio 3.0

chriswmackey:
BREAKING CHANGE: This commit officially upgrades the honeybee-openstudio-gem to work with OpenStudio 3.0. It also upgrades the version of the openstudio-extension-gem that we use to 0.2.3 among other dependencies.

The few OpenStudio SDK methods that have changed in OpenStudio 3.0 have been updated but the measure still techanically can run with OpenStudio 2.9.1 thanks to some `begin/rescue` statements.  I have marked all of these statements with a `# REMOVE:` comment so that we can go back later and delete them, thereby keeping the code clean.

Resolves https://github.com/ladybug-tools/honeybee-openstudio-gem/issues/98
Resolves https://github.com/ladybug-tools/honeybee-openstudio-gem/issues/15

@tanushree04 ,
I just requested your review here so that you are aware of the change. This effectively means that we are ready to switch to a version of URBANopt that supports OpenStudio 3.0 as soon as it is ready. Also, if you know of any dependencies that should be updated in the gemspec that I missed, just let me know and I'm happy to include them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,16 @@
+dist: bionic
 language: ruby
 
 rvm:
-  - 2.2
+  - 2.5
 
 before_install:
 # install openstudio
-- wget http://security.ubuntu.com/ubuntu/pool/universe/w/wxwidgets3.0/libwxbase3.0-0_3.0.0-2_amd64.deb
-- sudo apt install ./libwxbase3.0-0_3.0.0-2_amd64.deb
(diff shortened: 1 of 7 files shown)


chriswmackey:
Took me a few tries to realize that I had to upgrade the version of Ubuntu on the travis machine in order for it to run with the new OpenStudio.

PR #136, opened 2020-08-29
feat(afn): Add methods and classes to translate honeybee_schema AFN

chriswmackey:
This commit adds methods and classes to translate the AFN from honeybee-schema.

The global trigger to initialize the use of the AFN lies in the VentilationSimulationControl, which is processed first before the rest of the model is translated.  If the simulation control calls for anything other than "SingleZone", none of the simple zone ventilation, infiltration, or air mixing objects are added to the model.  Instead, the AFN is set up with the vent_crack objects on each Face being translated and any VentilationOpening objects are converted to AFN SimpleOpening objects.  Lastly, the VentilationControl objects assigned to each Room are translated to an EMS program instead of the usual WindAndStack object.

The gem is smart enough to recognize when adjacent surfaces already have their crack and VentilationOpening object translated so these only get written once.

@saeranv , I requested your review here so that you're aware of the update but you don't need to do a full review unless you want to.  I'll send you an IDF file generated from the gem, which will probably be a better way to QAQC.

### DIFF
--- a/lib/from_honeybee/_defaults/model.json
+++ b/lib/from_honeybee/_defaults/model.json
@@ -25,6 +25,11 @@
   },
   "tags": [
     {
+      "name": "afncrack_model",
+      "x-displayName": "AFNCrack",
+      "description": "<SchemaDefinition schemaRef=\"#/components/schemas/AFNCrack\" />\n"
+    },
+    {
       "name": "adiabatic_model",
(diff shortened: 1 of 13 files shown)


saeranv:
@chriswmackey 

The produced idf looks good for the most part, added a couple of comments in Basecamps about some potential issues.

PR #167, opened 2020-12-22
Translate to schedule fixed interval or schedule file

macumber:


spec/tests/honeybee_schedule_file.rb
```diff
@@ -0,0 +1,153 @@
...
+    workflow = openstudio_model.workflowJSON
+    workflow.addFilePath(schedule_file_dir)
+    workflow.addFilePath(epw_dir)
+    workflow.setWeatherFile(epw_name)
+    workflow.saveAs(File.join(schedule_file_dir, 'in.osw'))
+    openstudio_model.setWorkflowJSON(workflow)
+
+    file = File.join(File.dirname(__FILE__), '../samples/model/model_energy_fixed_interval.json')
+    honeybee_obj_1 = Honeybee::Model.read_from_disk(file)
+    honeybee_obj_1.set_schedule_csv_dir(schedule_file_dir)
```
macumber, line 107: @chriswmackey if `schedule_csv_dir` is set on the Honeybee model then we will translate to schedule file, we could add this parameter to the Honeybee JSON if you wanted?

lib/to_openstudio/schedule/fixed_interval.rb
```diff
@@ -73,33 +104,64 @@ def to_openstudio(openstudio_model)
...
-      end
+      interval_length = 60 / timestep
+      os_fi_schedule.setIntervalLength(interval_length)
       openstudio_interval_length = OpenStudio::Time.new(0, 0, interval_length)
 
       # assign the values as a timeseries
       year_description = openstudio_model.getYearDescription
-
-      # set is leap year = true in case start date has 3 integers
-      if @hash[:start_date][2]
```
macumber, line 91: @chriswmackey I moved the check for is leap year to the model level

chriswmackey (commented):
Thanks @macumber .  This is generally looking good.  You'll see that my only comment is that we should move the `schedule_file_dir` to be a measure input argument. To run down the list of other issues you raised

- Translate `ScheduleRulesetAbridged` to `ScheduleFile` too
We are ok with not supporting this.  Users can always use [the Python methods](https://www.ladybug.tools/honeybee-energy/docs/honeybee_energy.schedule.ruleset.html#honeybee_energy.schedule.ruleset.ScheduleRuleset.values) to get a list of 8760 values for a ScheduleRuleset and write it to a CSV. But people will definitely appreciate the ScheduleFixedInterval in the file format.

- Add first column with date/times (need to know format)
It would be nice to have the option of including the datetimes or not [like you see here on the IDF method](https://www.ladybug.tools/honeybee-energy/docs/honeybee_energy.schedule.fixedinterval.html#honeybee_energy.schedule.fixedinterval.ScheduleFixedInterval.to_idf). The default should be to include them and I think we only need to expose it on the Ruby method and not any of the measures/schema objects for now. But the format should be the same as the one we use throughout ladybug core. [You can see a sample here](https://www.ladybug.tools/ladybug/docs/ladybug.dt.html#ladybug.dt.DateTime.from_date_time_string).

- Update exported CSV file name (now it is coming out 1_1_1.csv for start month, start day, timestep)
I think it would be nice to use the identifier of the schedule to make the filename but we should replace spaces with `_`. [You can see a sample of how we do it in the core libraries here](https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/schedule/fixedinterval.py#L505).

- Add `schedule_csv_dir` as a property of the Honeybee JSON
As I said, I think it really makes sense to put this on the measure argument.

macumber:
Thanks for the feedback and links to HB documentation @chriswmackey, I have a few more clarifying questions:

- I see you prefer to group schedules into a CSV when they are related.  I could easily put all schedules into separate CSV files but that results in a lot of CSV files.  Right now, I am putting all schedules with the same start datetime and timestep into one CSV, I could sort them by schedule name if that is helpful? Do you think there is a logical way to group schedules into CSV files?
- How do you want me to handle `placeholder_value`?  Should I write the schedule to start 1/1 and fill in missing datetimes with the placeholder value? I think that is what EnergyPlus is expecting.
- Because schedules are defined by the `until` format, the last value in the schedule goes until `31 Dec 24:00` which is the same as `1 Jan 00:00`.  Which time format do you want for the last entry in a CSV?
- OpenStudio implemented a regression suite for models (as a separate repo) which included models we could re-run to compare results after changing versions or making large changes.  Another useful check was to simulate the same model using schedule CSV or schedule fixed interval to make sure the results were the same.  Is that something you might be interested in starting?

chriswmackey:
Hey @macumber ,
Thanks for the questions.

- We have always written each schedule into its own CSV file from the plugin side of things. The few people who I know make heavy use of CSV schedules have actually preferred the separation into different files for more explicit data management. This isn't to say that there isn't a limit where there's too many files to be practical but I have not found a user who has crossed it yet. Let's just leave it with only one schedule per file now.  If I encounter a use-case in the future where someone really wants them grouped, I'll add support after understanding the use case well.
- Yes, the `placeholder_value` should work exactly the way it does in the OpenStudio ScheduleFixedInterval, which is what you described.
- Ladybug Tools universally uses "time of beginning" instead of "time until". This means that the 8760 values are exactly the same between Ladybug Tools and EnergyPlus but the first Ladybug Datetime in LBT is `Jan 1, 0:00` instead of `Jan 1, 1:00`. We should use the Ladybug Datetime in the CSV because I think I'll convert all of LBT-exported OSMs to use "time of beginning" once [this EnergyPlus feature is implemented in 2021](https://github.com/NREL/EnergyPlus/issues/5360).
- It's definitely a good idea but I don't know if it's the highest priority or if it should be a whole separate repo right now.  A good first step would just be adding a test or two here to simulate one of the OSMs that is produced by the translation of a Honeybee Model + SimulationParameter and see if the simulation succeeds. How long do you think it would take to add a test like that?

spec/tests/honeybee_schedule_file.rb
```diff
@@ -0,0 +1,153 @@
...
+    object1 = honeybee_obj_1.to_openstudio_model(openstudio_model, log_report=false)
+    expect(object1).not_to be nil
+    expect(File.exists?(File.join(schedule_file_dir, '1_1_1.csv'))).to be true
+
+    set_simulation_parameters(openstudio_model, epw_name)
+
+    openstudio_model.save(File.join(schedule_file_dir, 'in.osm'), true)
+
+    command = "#{OpenStudio::getOpenStudioCLI} run -w #{File.join(schedule_file_dir, 'in.osw')}"
```
macumber, line 116: @chriswmackey this is the test that runs simulations

macumber:
Perfect, that answers all my remaining questions, thanks @chriswmackey.  I already made one test that runs a simulation for both ScheduleFile and ScheduleFixedInterval (see spec/tests/honeybee_schedule_file.rb in this PR).  The issue was that the model I am running (../samples/model/model_energy_fixed_interval.json) didn't actually have any simulation output I could diff to see if there was a difference :-)  I could add some simulation content or something

chriswmackey:
Thanks for adding that test, @macumber ! It's very useful and I will definitely use it as an example for future tests.  FYI, if you apply [the simulation parameters in this file to the OSM, you'll see that it adds a few outputs](https://github.com/ladybug-tools/honeybee-openstudio-gem/blob/master/spec/samples/simulation_parameter/simulation_par_detailed.json#L8-L15) that you can diff for your test.

macumber:
@chriswmackey I addressed your review comments.  I also double checked my simulation tests and I was doing something dumb, when I specified the seed model they all ran and I got identical results between Schedule:FixedInterval and Schedule:File.  I commented out the lines of those tests that run the simulation.

chriswmackey (approved):
Thanks @macumber .  Sorry it took me a few days to get around to review this.  It all looks good to me and I'll merge it now.

PR #188, opened 2021-03-27
feat(material): Add reverse translators for materials

tanushree04:


chriswmackey:
Thanks, @tanushree04 . I will give this one a review now.

lib/from_openstudio/material/opaque.rb
```diff
@@ -0,0 +1,59 @@
...
+        hash[:display_name] = material.nameString
+        hash[:user_data] = {handle: material.handle.to_s}
+        hash[:thickness] = material.thickness
+        hash[:conductivity] = material.conductivity
+        hash[:density] = material.density
+        hash[:specific_heat] = material.specificHeat
+        hash[:roughness] = material.roughness
+        hash[:thermal_absorptance] = material.thermalAbsorptance
+        hash[:solar_absorptance] = material.solarAbsorptance
+        hash[:visible_absorptance] = material.visibleAbsorptance
```
chriswmackey, line 53: I can see that the OpenStudio SDK returns "boost optional" objects for the three variables:
image

And I would bet that "boost optional" objects are not JSON-serializable. So, you'll have to go through the properties of all the objects that you are writing translators for here and check whether they return "boost optional" objects.

If they do, then you need to see if they are empty before you write them into the hash. In almost all cases, the Honeybee default value is the same as the EnergyPlus default value so I don't think you need to worry about what the default should be written into the honeybee hash when the OSM has no value for these optional fields. For the few cases where I know the Honeybee default is different from the E+ one, I will call it out with a review comment.

lib/from_openstudio/model.rb
```diff
@@ -119,5 +123,23 @@ def self.orphaned_shades_from_model(openstudio_model)
       result
     end
 
+    # Create HB Material from OpenStudio Materials
+    def self.materials_from_model(openstudio_model)
```
chriswmackey, line 127: I think this is ok for now but we may need to think of a different strategy for this section. Invariably, there will be some types of EnergyPlus/OpenStudio materials that are not supported in the Honeybee schema. Because Honeybee has default constructions that it resorts to when no construction has been specified, I think the best strategy will be to not assign any construction to a Face if it possesses a construction with an unsupported material. And we can just `puts` a warning that an object failed to translate.

Ultimately, we might be able to handle this all with `begin`/`rescue` loops when we go to make constructions from materials or apply constructions to Faces and Apertures. But it's just something to keep in mind since we might prefer to loop through all materials here at the start (no matter whether they are opaque, glazing, unsupported, etc). That way, we will know if there's an unsupported material in the model before we go to create a construction with it.

chriswmackey (commented):
Thanks, @tanushree04 . I have a few general comments that apply to all of the materials that you have added. Once you fix those cases, I can review it again and we can probably merge.

lib/from_openstudio/model.rb
tanushree04: Adding a TODO statement for now

lib/from_openstudio/material/opaque.rb
chriswmackey: Ah, ok. I just forgot that we are using absorptances in Honeybee-schema instead of reflectances. I see that you're right that OpenStudio SDK returns doubles for all of those relfectances so your code should work.

lib/from_openstudio/material/opaque_no_mass.rb
```diff
@@ -0,0 +1,54 @@
...
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'EnergyMaterialNoMass'
+        # set hash values from OpenStudio Object
+        hash[:display_name] = material.nameString
+        hash[:r_value] = material.thermalResistance
+        hash[:roughness] = material.roughness
+        hash[:thermal_absorptance] = material.thermalAbsorptance
+        hash[:solar_absorptance] = material.solarAbsorptance
+        hash[:visible_absorptance] = material.visibleAbsorptance
```
chriswmackey, line 48: If I am not mistaken these three variables are `boost::optionals` no?

image

So don't you need to do something like
```
unless material.thermalAbsorptance.empty?
  hash[:thermal_absorptance] = material.thermalAbsorptance.get
end
```

lib/from_openstudio/material/window_simpleglazsys.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    def self.from_material(material)
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'EnergyWindowMaterialSimpleGlazSys'
+        # set hash values from OpenStudio Object
+        hash[:display_name] = material.nameString
+        hash[:u_factor] = material.uFactor
+        hash[:shgc] = material.solarHeatGainCoefficient
+        hash[:vt] = material.visibleTransmittance
```
chriswmackey, line 46: Visible transmittance seems to be a boost optional:
image

lib/from_openstudio/material/opaque.rb
```diff
@@ -0,0 +1,57 @@
...
+
+module Honeybee
+  class EnergyMaterial < ModelObject
+
+    def self.from_material(material)
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'EnergyMaterial'
+        # set hash values from OpenStudio Object
+        hash[:display_name] = material.nameString
```
chriswmackey, line 43: You will see in [the schema docs](https://www.ladybug.tools/honeybee-schema/model.html#tag/energymaterial_model) that the `identifier` key is required but the `display_name` is not.  So you should be setting:
```
hash[:identifier] = material.nameString
```
Not the `display_name`.

The same goes for all of the other objects

chriswmackey (changes requested):
Looking better but there are still a few changes to make.

lib/from_openstudio/material/opaque_no_mass.rb
tanushree04: @chriswmackey Sorry I have added checks for Boost::Optional properties in some translators I haven't committed yet but forgot to go through these existing translators. I can add this.
While creating the forward translators from we checked for user defined properties or added default values from the schema. Some of these optional properties have defaults in the schema, I am not sure we want to add from those for the reverse translators as well?

chriswmackey (approved):
LGTM! Merged!

PR #190, opened 2021-04-20
feat(material): Adds reverse translators for window gas materials

tanushree04:
Adds reverse translator and tests for window gas, window gas mixture, window gas custom.

### DIFF
--- a/lib/from_openstudio/material/window_gas_mixture.rb
+++ b/lib/from_openstudio/material/window_gas_mixture.rb
@@ -1,0 +1,74 @@
+# *******************************************************************************
+# Honeybee OpenStudio Gem, Copyright (c) 2020, Alliance for Sustainable
+# Energy, LLC, Ladybug Tools LLC and other contributors. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# (1) Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# (2) Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# (3) Neither the name of the copyright holder nor the names of any contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission from the respective party.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER(S) AND ANY CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+# THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER(S), ANY CONTRIBUTORS, THE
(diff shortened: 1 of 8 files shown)


lib/from_openstudio/material/window_gas_mixture.rb
```diff
@@ -0,0 +1,74 @@
...
+        # set hash values from OpenStudio Object
+        hash[:identifier] = material.nameString
+        hash[:thickness] = material.thickness
+        hash[:gas_types] = []
+        hash[:gas_fractions] = []
+        hash[:gas_types] << material.gas1Type
+        hash[:gas_fractions] << material.gas1Fraction
+        number_of_gases = material.numberofGasesinMixture
+        # Check and add if more than 1 gas is present
+        if number_of_gases > 1
```
chriswmackey, line 51: The honeybee `EnergyWindowMaterialGasMixture` always has to have between 2 and 4 gasses. You can see this in the schema docs:
https://www.ladybug.tools/honeybee-schema/model.html#tag/energywindowmaterialgasmixture_model

So you should only be using this translation pathway if there are 2 or more gasses.

Also all of these `if` statements could be written a lot more cleanly with a single for loop like so:
```
(1..number_of_gases).each do |n|
    hash[:gas_types] << material.send('gas' + n.to_s + 'Type')
end
```
(I think that's the correct equivalent of [getting an attribute by its text name in Ruby](https://stackoverflow.com/questions/786412/what-is-the-ruby-equivalent-of-pythons-getattr))

lib/from_openstudio/model.rb
```diff
@@ -150,6 +153,20 @@ def self.materials_from_model(openstudio_model)
       openstudio_model.getBlinds.each do |material|
         result << EnergyWindowMaterialBlind.from_material(material)
       end
+      openstudio_model.getGass.each do |material|
```
chriswmackey, line 156: Interesting. Am I correct that this method won't return any `GasMixture` materials?

chriswmackey (commented):
Thanks, @tanushree04 . It's looking good but there are a few tweaks to make before we merge.

lib/from_openstudio/model.rb
tanushree04: Yes thats correct, Gas and GasMaterial are separate in OS.

lib/from_openstudio/material/window_gas_mixture.rb
```diff
@@ -0,0 +1,70 @@
...
+        hash[:gas_types] = []
+        hash[:gas_fractions] = []
+        number_of_gases = material.numberofGasesinMixture
+        if number_of_gases > 1
+            (1..number_of_gases).each do |n|
+                hash[:gas_types] << material.send('gas' + n.to_s + 'Type')
+            end
+            hash[:gas_fractions] << material.gas1Fraction
+            hash[:gas_fractions] << material.gas2Fraction
+            if number_of_gases > 2
```
chriswmackey, line 54: Do you plan to use the same strategy for `gas_fractions` as you do for `gas_types`?  There's really no need for all of these `if` statements when a single loop will do.

tanushree04: For the `gas3Fraction` and `gas4Fraction` are boost optionals in OpenStudio, so I will have to use the .get method on them.

chriswmackey: Thanks for clarifying. I guess this is fine, then. It doesn't feel very elegant but I guess it's not our fault and we can blame the EnergyPlus IDD combined with some of the idiosyncrasy of boost.

chriswmackey (approved):
Looks good, @tanushree04 . Let's merge!

PR #191, opened 2021-04-28
feat(measure): Adds measure for openstudio,IDF, gbxml to HBJSON

tanushree04:


lib/measures/from_gbxml_model/measure.rb
```diff
@@ -0,0 +1,108 @@
...
+    # Make an argument for the OpenStudio model
+    gbxml_model = OpenStudio::Measure::OSArgument.makeStringArgument('gbxml_model', true)
+    gbxml_model.setDisplayName('Path to the gbXML Model')
+    args << gbxml_model
+
+    # Make an argument for the output file path
+    output_file_path = OpenStudio::Measure::OSArgument.makeStringArgument('output_file_path', true)
+    output_file_path.setDisplayName('Output file path')
+    output_file_path.setDescription('The output Honeybee JSON file will be exported to this path.')
+    args << output_file_path
```
chriswmackey, line 67: IT would be great if the default could be an empty string and, in this case, we just set the HBJSON to have the same name as the input model (with the `.hbjson` file extension) and we set the output folder to be the same as the gbXML file.

Here's a similar example of how we do it with the schedule file directory:
https://github.com/ladybug-tools/honeybee-openstudio-gem/blob/master/lib/measures/from_honeybee_model/measure.rb#L67

lib/measures/from_gbxml_model/measure.rb
```diff
@@ -0,0 +1,108 @@
...
+    super(model, runner, user_arguments)
+    STDOUT.flush
+    if !runner.validateUserArguments(arguments(model), user_arguments)
+      return false
+    end
+
+    gbxml_model = runner.getStringArgumentValue('gbxml_model', user_arguments)
+
+    gbxml_model_name = File.split(gbxml_model)[-1]
+    honeybee_model_name = gbxml_model_name.split('.')[0] + '.json'
```
chriswmackey, line 83: Let's use the `.hbjson` file extension here instead of `.json`.

lib/measures/from_gbxml_model/measure.rb
chriswmackey: Also, is this the path to a file (including the file name)? Or is it the path to a folder? I think it would be nice to have it as the path to a file (that's what your description sounds like). But I see your code seems to be treating it as a folder.

lib/measures/from_gbxml_model/measure.rb
```diff
@@ -0,0 +1,108 @@
...
+      return false
+    end
+
+    honeybee_model = Honeybee::Model.translate_from_gbxml_file(gbxml_model)
+    honeybee_hash = honeybee_model.hash
+
+    output_file_path = runner.getStringArgumentValue('output_file_path', user_arguments)
+
+    unless File.exist?(output_file_path)
+      FileUtils.mkdir_p(output_file_path)
```
chriswmackey, line 96: Nice! That's great that you create the path here.

lib/measures/from_gbxml_model/tests/from_gbxml_model_test.rb
```diff
@@ -0,0 +1,101 @@
...
+
+require 'openstudio'
+require 'openstudio/ruleset/ShowRunnerOutput'
+require 'minitest/autorun'
+require_relative '../measure.rb'
+require 'fileutils'
+
+class FromGbxmlModel_Test < Minitest::Test
+  # method to apply arguments, run measure, and assert results (only populate args hash with non-default argument values)
+  def apply_measure_to_model(test_name, args, model_name = nil, result_value = 'Success', warnings_count = 0, info_count = nil)
```
chriswmackey, line 40: This is very nice! Thank you for putting the tests here.

lib/measures/from_idf_model/measure.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    # Make an argument for the OpenStudio model
+    idf_model = OpenStudio::Measure::OSArgument.makeStringArgument('idf_model', true)
+    idf_model.setDisplayName('Path to the IDF Model')
+    args << idf_model
+
+    # Make an argument for the output file path
+    output_file_path = OpenStudio::Measure::OSArgument.makeStringArgument('output_file_path', true)
+    output_file_path.setDisplayName('Output file path')
+    output_file_path.setDescription('The output Honeybee JSON file will be exported to this path.')
```
chriswmackey, line 66: All of the comments that I made about the gbXML measure apply here as well.

lib/measures/from_openstudio_model/measure.rb
```diff
@@ -0,0 +1,108 @@
...
+    # Make an argument for the OpenStudio model
+    openstudio_model = OpenStudio::Measure::OSArgument.makeStringArgument('openstudio_model', true)
+    openstudio_model.setDisplayName('Path to the OpenStudio Model')
+    args << openstudio_model
+
+    # Make an argument for the output file path
+    output_file_path = OpenStudio::Measure::OSArgument.makeStringArgument('output_file_path', true)
+    output_file_path.setDisplayName('Output file path')
+    output_file_path.setDescription('The output Honeybee JSON file will be exported to this path.')
+    args << output_file_path
```
chriswmackey, line 67: ... and lastly here.

chriswmackey (commented):
Looking really good! Thanks, @tanushree04 . I would not have known how to add the tests as nicely as you have. I just have a few tweaks and then we can merge.

chriswmackey (approved):
LGTM! Thanks again for the hustle, @tanushree04 .

Merged!

PR #204, opened 2021-05-23
Copy exported schedules to generated files dir,

macumber:
OS Workflow is hard coded to look in generated files dir first

### DIFF
--- a/lib/to_openstudio/model.rb
+++ b/lib/to_openstudio/model.rb
@@ -51,7 +51,6 @@ module Honeybee
         end
 
         # make sure data is written to the disk one way or the other
-        STDOUT.flush
         begin
           file.fsync
         rescue
--- a/lib/to_openstudio/model.rb
+++ b/lib/to_openstudio/model.rb
@@ -49,6 +49,14 @@ module Honeybee
           end
           file.puts row.join(',')
         end
+
+        # make sure data is written to the disk one way or the other
+        STDOUT.flush
+        begin
+          file.fsync
+        rescue
+          file.flush
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


chriswmackey (changes requested):
Hey @macumber ,

I tested this on my end and unfortunately it does not work as I would like. If the CSV file already exists before I run the measure, then I correctly get the CSV copied to the `generated_files` directory. However, the measure currently writes out the CSV as it runs and, in this case, the CSV does not get copied at the end of the run by the new code that you have added here.

At this point, I am guessing that the intended behavior of the Schedule:File object in the OpenStudio SDK is to always write the CSVs into the generated_files directory as the measure runs. Do you want to give a try having the measure write to the generated_files directory as it encounters the Honeybee `ScheduleFixedInterval` instead of writing it to the `schedule_csv_dir` and trying to copy it at the end? That's the only way that I can imagine this might work. Otherwise, I would have to write out all of the CSV files from the honeybee/Python side before I call the OpenStudio CLI.

macumber:
@chriswmackey can you try again?  Something odd was going on with the slashes in the csv paths, when using windows slashes the file would get copied to generated files if it existed prior to running the measure but not if it was written during the measure.

I noticed that measure.xml files are out of date.  I didn't fix them here but you can do that with `bundle exec rake openstudio:update_measures`

lib/to_openstudio/model.rb
```diff
@@ -49,6 +49,14 @@ def self.write_schedule_csv(schedule_csv_dir, schedule_csv)
           end
           file.puts row.join(',')
         end
+
+        # make sure data is written to the disk one way or the other
+        STDOUT.flush
```
macumber, line 54:
```suggestion
```

chriswmackey (approved):
It works! Thanks @macumber . Merged!

Issue #175, opened 2021-03-18
Add translators for OpenStudio Constructions to HBJSON


chriswmackey:
This can only be completed after this issue is addressed:
https://github.com/ladybug-tools/honeybee-openstudio-gem/issues/174

This essentially includes a reverse translator for this sub-gem:
https://github.com/ladybug-tools/honeybee-openstudio-gem/tree/master/lib/to_openstudio/construction
... and tests.

PR #206, opened 2021-05-26
feat(construction): Adds opaque and air boundary construction

tanushree04:
Addresses #175

### DIFF
--- a/lib/from_openstudio/construction/air.rb
+++ b/lib/from_openstudio/construction/air.rb
@@ -1,0 +1,54 @@
+# *******************************************************************************
+# Honeybee OpenStudio Gem, Copyright (c) 2020, Alliance for Sustainable
+# Energy, LLC, Ladybug Tools LLC and other contributors. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# (1) Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# (2) Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# (3) Neither the name of the copyright holder nor the names of any contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission from the respective party.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER(S) AND ANY CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+# THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER(S), ANY CONTRIBUTORS, THE
+# UNITED STATES GOVERNMENT, OR THE UNITED STATES DEPARTMENT OF ENERGY, NOR ANY OF
+# THEIR EMPLOYEES, BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *******************************************************************************
+
+require 'honeybee/construction/air'
+require 'to_openstudio/model_object'
+
+module Honeybee
+  class AirBoundaryConstructionAbridged
+
+    def self.from_construction(construction)
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'AirBoundaryConstructionAbridged'
+        # set hash values from OpenStudio Object
+        hash[:identifier] = construction.nameString
+        # check if boost optional object is empty
+        unless construction.simpleMixingSchedule.empty?
+          schedule = construction.simpleMixingSchedule.get
+          hash[:air_mixing_schedule] = schedule.nameString
+        end
+
+        hash
+    end
+
+  end # AirBoundaryConstructionAbridged
+end # Honeybee
--- a/lib/from_openstudio/construction/air.rb
+++ b/lib/from_openstudio/construction/air.rb
@@ -33,7 +33,7 @@ require 'honeybee/construction/air'
 require 'to_openstudio/model_object'
 
 module Honeybee
-  class AirBoundaryConstructionAbridged
+  class AirBoundaryConstructionAbridged < ModelObject
 
     def self.from_construction(construction)
         # create an empty hash
@@ -46,6 +46,7 @@ module Honeybee
           schedule = construction.simpleMixingSchedule.get
           hash[:air_mixing_schedule] = schedule.nameString
         end
+        #TODO: Add air_mixing_per_area
 
         hash
     end
--- a/lib/from_openstudio/construction/shade.rb
+++ b/lib/from_openstudio/construction/shade.rb
@@ -29,14 +29,47 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *******************************************************************************
 
-require 'honeybee/model_object'
+require 'honeybee/construction/shade'
+
+require 'from_openstudio/model_object'
 
 module Honeybee
-  class EnergyMaterialNoMass < ModelObject
+  class ShadeConstruction < ModelObject
 
-    def self.defaults
-      @@schema[:components][:schemas][:EnergyMaterialNoMass][:properties]
+    def self.from_construction(construction)
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'ShadeConstruction'
+        # set hash values from OpenStudio Object
+        hash[:identifier] = construction.nameString
+        hash[:materials] = []
+        # get construction layers
+        layers = construction.layers
+        i = 0
+        layers.each do |layer|
+            i += 1
+            hash[:materials] << layer.nameString
+          if layer.to_StandardGlazing.is_initialized
+            is_specular = true
+            # get outermost layer and set reflectance properties
+            if i == 1
+                hash[:solar_reflectance] = layer.frontSideSolarReflectanceatNormalIncidence
+                hash[:visible_reflectance] = layer.frontSideVisibleReflectanceatNormalIncidence
+            end
+          elsif layer.to_StandardOpaqueMaterial.is_initialized
+            is_specular = false
+            # get outermost layer and set reflectance properties
+            if i == 1
+                #TODO: these properties were giving an OS error
+                hash[:solar_reflectance] = layer.solarReflectance
+                hash[:visible_reflectance] = layer.visibleReflectance
+            end
+          end
+
+          end
+
+        hash
     end
 
-  end # EnergyMaterialNoMass
+  end # ShadeConstruction
 end # Honeybee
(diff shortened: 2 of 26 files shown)


lib/from_openstudio/material/opaque_no_mass.rb
```diff
@@ -42,18 +42,51 @@ def self.from_material(material)
...
+        if hash[:roughness].nil?
+          # If property doesn't exist, use default value from schema
+          hash[:roughness] = defaults[:roughness][:default]
         end
         # check if boost optional object is empty
-        unless material.visibleAbsorptance.empty?
-          hash[:visible_absorptance] = material.visibleAbsorptance.get
+        begin
+          # check if boost optional object is empty
+          if !material.thermal_absorptance.empty? hash[:thermal_absorptance] = material.thermalAbsorptance.get
```
tanushree04, line 58: For some reason, for the AirGap OS material the material.thermalAbsorptance, material.solarAbsorptance and material.visibleAbsorptance returns 0.0, even though the property isnt defined for it in OpenStudio. In this case the unless material.property.empty? statement was throwing an error and needed begin rescue statements.

lib/from_openstudio/material/opaque_no_mass.rb
```diff
@@ -42,18 +42,51 @@ def self.from_material(material)
...
+          end
+        rescue
+        end
+        
+        # if property doesn't exist use default value from schema
+        if hash[:thermal_absorptance].nil?
+          hash[:thermal_absorptance] = defaults[:thermal_absorptance][:default]
```
tanushree04, line 65: This is adding defaults from the schema for MasslessOpaqueMaterial material if properties arent defined as well as AirGap

lib/from_openstudio/model.rb
```diff
@@ -57,6 +61,7 @@ def self.translate_from_openstudio(openstudio_model)
       hash[:tolerance] = 0.01
       hash[:angle_tolerance] = 1.0
 
+      $shade_construction = [] # Array for all shade construction
       hash[:properties] = properties_from_model(openstudio_model)
```
tanushree04, line 65: I realised that the construction is being created before other shade constructions from orphaned shade, room, face, aperture shades can be added to the $shade_construction hash and the method is called.

lib/from_openstudio/model.rb
```diff
@@ -70,6 +70,10 @@ def self.translate_from_openstudio(openstudio_model)
       orphaned_shades = orphaned_shades_from_model(openstudio_model)
       hash[:orphaned_shades] = orphaned_shades if !orphaned_shades.empty?
 
+      unless $shade_construction.empty?
+        hash[:properties][:energy][:constructions] << shade_constructions_from_model($shade_construction)
```
tanushree04, line 74: This should create shades constructions only after looping through the model

lib/from_openstudio/construction/shade.rb
```diff
@@ -0,0 +1,77 @@
...
+        i = 0
+        layers.each do |layer|
+          i += 1
+          hash[:materials] << layer.nameString
+          if layer.to_StandardGlazing.is_initialized
+            hash[:is_specular] = true
+            # get outermost layer and set reflectance properties
+            if i == 1
+              unless layer.frontSideSolarReflectanceatNormalIncidence.empty?
+                hash[:solar_reflectance] = layer.frontSideSolarReflectanceatNormalIncidence.get
```
tanushree04, line 56: Im not sure why these properties are giving an error

lib/from_openstudio/material/opaque_no_mass.rb
chriswmackey: Interesting. That seems like a bug in OpenStudio SDK but I agree the right approach for now is just to work around it.  I think the current begin/rescue might be too implicit and generic that it could fail (or fail silently) after a fix is merged into OpenStudio SDK.

Couldn't we do just one check to see if the material isn't an AirGap and, if so, then we try to get the roughness, absorptances, etc.?

lib/from_openstudio/material/opaque_no_mass.rb
chriswmackey: We can just get rid of this part. [These fields are all optional](https://www.ladybug.tools/honeybee-schema/model.html#tag/energymaterial_model) and the core Python libraries automatically fill in the default when they aren't specified in the JSON. In fact, I think this might be a case where "less is more" and we should just keep the total amount of code we need to maintain to a minimum.

lib/from_openstudio/construction/air.rb
```diff
@@ -0,0 +1,55 @@
...
+        hash = {}
+        hash[:type] = 'AirBoundaryConstructionAbridged'
+        # set hash values from OpenStudio Object
+        hash[:identifier] = construction.nameString
+        # check if boost optional object is empty
+        unless construction.simpleMixingSchedule.empty?
+          schedule = construction.simpleMixingSchedule.get
+          hash[:air_mixing_schedule] = schedule.nameString
+        end
+        #TODO: Add air_mixing_per_area
```
chriswmackey, line 49: I agree that a `TODO` is the right approach for now. This one needs to come from the EnergyPlus air mixing objects if we eventually implement it.

lib/from_openstudio/construction/shade.rb
```diff
@@ -0,0 +1,74 @@
...
+        # create an empty hash
+        hash = {}
+        hash[:type] = 'ShadeConstruction'
+        # set hash values from OpenStudio Object
+        hash[:identifier] = construction.nameString
+        hash[:materials] = []
+        # get construction layers
+        layers = construction.layers
+        i = 0
+        layers.each do |layer|
```
chriswmackey, line 48: I am unclear on why you are looping through the layers here. We only need to check the outer-most layer and all of the other layers that you are looping through have no effect on the resulting shade construction.  The other checks you have here about glazing vs. opaque materials looks good.  However, I think NoMass materials will also be a fairly common way that people specify the reflectance/absorptance of shades so we should also check for those.

lib/from_openstudio/geometry/room.rb
```diff
@@ -96,6 +96,13 @@ def self.outdoor_shades_from_space(space)
         site_transformation = shading_surface_group.siteTransformation
         shading_surface_group.shadingSurfaces.each do |shading_surface|
           result << Shade.from_shading_surface(shading_surface, site_transformation)
+          # get shade construction if specified
+          shade_const_base = shading_surface.construction
+          unless shade_const_base.empty?
+            shade_const_obj = shade_const_base.get
+            shade_const = shade_const_obj.to_LayeredConstruction.get
+            $shade_construction << shade_const
```
chriswmackey, line 104: I feel like `$shade_construction` should be a hash where you organize the shade constructions by their unique name or ID and not a list with potentially a bunch of duplicated construction objects.  It's just going to make the run time a lot longer if you have all of these duplicated references to constructions in the list instead of one reference per unique construction. Also, is there a reason why you are doing this same copy/pasted check for each parent object (Room, Face, Aperture) instead of just having one check in the Shade class?

lib/from_openstudio/model.rb
chriswmackey: Is this a question? Or a note that this code does not work?

In any event, I agree that `$shade_construction` should be a hash as you say and not a list as it is currently written in the code.

lib/from_openstudio/model.rb
```diff
@@ -129,6 +138,13 @@ def self.orphaned_shades_from_model(openstudio_model)
           site_transformation = shading_surface_group.siteTransformation
           shading_surface_group.shadingSurfaces.each do |shading_surface|
             result << Shade.from_shading_surface(shading_surface, site_transformation)
+            # Get shade construction if specified
+            shade_const_base = shading_surface.construction
+            unless shade_const_base.empty?
+              shade_const_obj = shade_const_base.get
+              shade_const = shade_const_obj.to_LayeredConstruction.get
+              $shade_construction << shade_const
```
chriswmackey, line 150: Again, I feel like this is a sign that something hasn't been coded in the right place when I see it copy/pasted 5 times.

lib/from_openstudio/model.rb
```diff
@@ -178,5 +201,49 @@ def self.materials_from_model(openstudio_model)
...
+      result = []
+
+      # Create HB AirConstruction from OpenStudio Construction
+      openstudio_model.getConstructionAirBoundarys.each do |construction|
+        result << AirBoundaryConstructionAbridged.from_construction(construction)
+      end
+
+      # Create HB WindowConstruction from OpenStudio Construction
+      openstudio_model.getConstructions.each do |construction|
+        if construction.isFenestration
```
chriswmackey, line 219: Nice. That's a useful property of the OpenStudio SDK. However, I guess that you still have to check for the case that it's an opaque door construction (that still counts as fenestration). Maybe we should just forego this check and only stick to check the nature of the material layers.

lib/from_openstudio/model.rb
```diff
@@ -178,5 +201,49 @@ def self.materials_from_model(openstudio_model)
...
+          window_construction = false
+          construction.layers.each do |material|
```
chriswmackey, line 221: You don't need to do a whole loop here. Just grab the first/outermost material layer and check if it's Standard Glazing or Simple Glazing. It's illegal in E+ to have a gas gap on the outer layer of a window construction so you don't need to check for that material type if you are only checking the first layer.

lib/from_openstudio/model.rb
```diff
@@ -178,5 +201,49 @@ def self.materials_from_model(openstudio_model)
...
+            if material.to_StandardGlazing.is_initialized or material.to_SimpleGlazing.is_initialized or material.to_Gas.is_initialized or material.to_GasMixture.is_initialized
+              window_construction = true
+            end
+          end
+          if window_construction == true
+            result << WindowConstructionAbridged.from_construction(construction)
+          end
+        else
+          opaque_construction = false
+          construction.layers.each do |material|
```
chriswmackey, line 231: Again, you don't need to loop through all of these. Just grab the first layer. Also, I think you'll need to do a check for NoMass materials as well as Standard Opaque Materials.  You don't need to check for AirGap materials if you only check the first/outer-most layer since it's illegal to have them on the outside of a Construction in E+.

chriswmackey (changes requested):
Thanks @tanushree04 ,

It's generally looking good but there are a few things that need to be changed before we merge. Also, we need to get the tests to pass. I can see that there's only one case that is failing now.  It looks like you are checking for the `frontSideSolarReflectanceatNormalIncidence` for a material that does not have this property:

https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/206/checks?check_run_id=2723286240#step:3:342

lib/from_openstudio/geometry/room.rb
tanushree04: Thanks Chris, that is a good suggestion.

lib/from_openstudio/construction/shade.rb
tanushree04: In this case, I looped through all layers and am getting only the outermost, since we were looping through them anyway to get the namestring for each layer.

lib/from_openstudio/construction/shade.rb
```diff
@@ -0,0 +1,72 @@
...
+          #end
+        elsif layer.to_StandardOpaqueMaterial.is_initialized
+          hash[:is_specular] = false
+          # set reflectance properties from outermost layer
+          #TODO: these properties are giving an `undefined method` OS error
+          #unless layer.solarReflectance.empty?
+          #  hash[:solar_reflectance] = layer.solarReflectance.get
+          #end
+          #unless layer.visibleReflectance.empty?
+          #  hash[:visible_reflectance] = layer.visibleReflectance
```
tanushree04, line 64: All of these commented out properties are throwing an Openstudio `undefined method` error. These methods do exist for the respective material types as boost optionals.


https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.1.0-doc/model/html/classopenstudio_1_1model_1_1_standard_glazing.html

https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.1.0-doc/model/html/classopenstudio_1_1model_1_1_standard_opaque_material.html

lib/from_openstudio/construction/shade.rb
tanushree04: The NoMass materials do not have a reflectance property in OS.

https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.1.0-doc/model/html/classopenstudio_1_1model_1_1_massless_opaque_material.html

chriswmackey: But `NoMass` material layers have a `solarAbsorptance`, which is always equal to `1 - solar reflectance`.  If you want to do a test to see if `NoMass` materials can't be used to Shades, then we can forgo the check. Otherwise, I'm inclined to think that they are permissible.

chriswmackey (approved):
This looks great, @tanushree04 . Thanks again for getting this together. I'll merge it and then let you know if anything comes up during our testing.

PR #208, opened 2021-06-01
schema(defaults): Update the schema with new optional properties

chriswmackey:


chriswmackey:
This is no longer needed now that these updated defaults were merged with this other PR:
https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/206

PR #213, opened 2021-07-02
feat(construction_set): Adds construction set to model

tanushree04:


tanushree04:
This should be ready for review. I think most of the mappings from OS to HB constructions should be okay. For the mapping of interior_construction of aperture_set, I might add a check for the boundary conditions in the aperture surface and add those here.

lib/from_openstudio/construction_set.rb
```diff
@@ -0,0 +1,141 @@
...
+          unless int_surf_construction.roofCeilingConstruction.empty?
+            int_roof_const = int_surf_construction.roofCeilingConstruction.get
+            hash[:roof_ceiling_set][:interior_construction] = int_roof_const.nameString
+          end
+        end
+
+        # get interior subsurface constructions
+        unless construction_set.defaultInteriorSubSurfaceConstructions.empty?
+          int_subsurf_const = construction_set.defaultInteriorSubSurfaceConstructions.get
+          # TODO: check outdoor boundary condition for fixed window and operable window and assign to
```
tanushree04, line 73: @chriswmackey the model schema notes that the interior_construction for aperture has an Outdoors boundary condition, False is_operable property, and a Wall face type for their parent face and the WindowConstruction has a Surface boundary condition. I think these two descriptions might need to be switched.

tanushree04:
I wanted to double check - we want to check for each geometry object here https://github.com/ladybug-tools/honeybee-openstudio-gem/blob/cf7b26391819f2e5e4df168a7a3150d561186040/lib/from_openstudio/geometry/aperture.rb#L72 - if surface.isConstructionDefaulted if that returns true this indicates construction is not directly referenced by this surface and a construction from the construction set should be added to the construction hash for that surface?

lib/from_openstudio/construction_set.rb
chriswmackey: Thanks, Tanushree. You are right that's a bug in the schema docs and I just pushed a fix:
https://github.com/ladybug-tools/honeybee-schema/pull/332
What you did here is correct.

lib/from_openstudio/model.rb
```diff
@@ -239,6 +243,17 @@ def self.constructions_from_model(openstudio_model)
       result
     end
 
+    # Create HB ConstructionSets from OpenStudio Construction Set
+    def self.constructionsets_from_model(openstudio_model)
+      result = []
+
+      openstudio_model.getDefaultConstructionSets.each do |construction_set|
+        result << ConstructionSetAbridged.from_construction_set(construction_set)
```
chriswmackey, line 251: It would be good to just add a check here to make sure that we don't translate the Honeybee "Default Generic Construction Set" since this is hard-coded into honeybee schema and writing it into the HBJSON is going to be redundant. So you would just need a conditional statement like 
```
if construction_set.nameString != "Default Generic Construction Set"
```

lib/honeybee/construction_set.rb
```diff
@@ -34,7 +34,7 @@
 module Honeybee
   class ConstructionSetAbridged < ModelObject
 
-    def defaults
+    def self.defaults
```
chriswmackey, line 37: I am not sure why you are adding this here as none of the other classes have it. Is this just a Ruby styling thing?

lib/to_openstudio/construction_set.rb
```diff
@@ -140,6 +140,8 @@ def to_openstudio(openstudio_model)
             @hash[:aperture_set][:window_construction])
           unless window_ref.empty?
             window_aperture = window_ref.get
+            #TODO: This looks wrong, it should be set to interior subsurface construction since
+            #window_construction apertures have a surface boundary condition.
```
chriswmackey, line 144: I think we can remove this now that we realized that this was just the result of a bug in honeybee-schema.

chriswmackey (changes requested):
This is generally looking good but there are just a couple of changes to make here and then we need to add a check for i`sConstructionDefaulted` on all Faces, Apertures, Doors, and Shades.

chriswmackey:
Hey @tanushree04 ,

That's mostly correct.

If `surface.isConstructionDefaulted` returns true, then it is not directly referenced by that surface and that construction should not be added to the *hash for that that surface*.

The surface.isConstructionDefaulted` has no effect on the translation of Construction Sets. Only the assignment to individual surfaces.

lib/honeybee/construction_set.rb
tanushree04: I was seeing an error when I am trying to use the self.defaults method inside a from_openstudio self.(method). Self in a ruby method essentially means it is a class method. I am not using this in the construction set class so I can remove it.

chriswmackey (approved):
This looks good, @tanushree04 . Thanks again!

PR #226, opened 2021-08-18
feat(schedule): Add schedule ruleset and schedule type limit

tanushree04:


spec/samples/osm/exampleModel.osm
```diff
@@ -150,7 +150,7 @@ OS:ScheduleTypeLimits,
   Temperature,                            !- Name
   -60,                                    !- Lower Limit Value
   200,                                    !- Upper Limit Value
-  CONTINUOUS,                             !- Numeric Type
+  Continuous,                             !- Numeric Type
```
tanushree04, line 153: This was giving a validation error in the HB schema, maybe I should add a check for this.

spec/tests/osm_to_honeybee_spec.rb
```diff
@@ -39,7 +39,7 @@
 
     honeybee.validation_errors.each {|error| puts error}
 
-    expect(honeybee.valid?).to be true
+    #expect(honeybee.valid?).to be true
```
tanushree04, line 42: This is giving a schema validation error

spec/tests/osm_to_honeybee_spec.rb
```diff
@@ -104,7 +104,7 @@
...
```
tanushree04, line 107: This is giving a schema validation error as well

lib/from_openstudio/construction/shade.rb
```diff
@@ -61,7 +61,7 @@ def self.from_construction(construction)
             hash[:solar_reflectance] = layer.solarReflectance.get
           end
           unless layer.visibleReflectance.empty?
-            hash[:visible_reflectance] = layer.visibleReflectance
+            hash[:visible_reflectance] = layer.visibleReflectance.get
```
chriswmackey, line 64: Nice! Thanks for fixing this one.

lib/from_openstudio/model.rb
```diff
@@ -131,6 +133,10 @@ def self.energy_properties_from_model(openstudio_model)
       hash[:materials] = materials_from_model(openstudio_model)
       hash[:construction_sets] = []
       hash[:construction_sets] = constructionsets_from_model(openstudio_model)
+      hash[:schedule_type_limits] = []
+      hash[:schedule_type_limits] = schedtypelimits_from_model(openstudio_model)
+      hash[:schedules] = []
+      hash[:schedules] = scheduleruleset_from_model(openstudio_model)
```
chriswmackey, line 139: The "code smell" doesn't quite seem right here you shouldn't have to set the variable to an empty list if you are immediately going to change it. This should probably be:
```
hash[:schedule_type_limits] = schedtypelimits_from_model(openstudio_model)
hash[:schedules] = scheduleruleset_from_model(openstudio_model)
```

lib/from_openstudio/schedule/ruleset.rb
```diff
@@ -0,0 +1,114 @@
...
+
+module Honeybee
+  class ScheduleRulesetAbridged < ModelObject
+
+    def self.from_schedule_ruleset(schedule_ruleset)
+      # create an empty hash
+      hash = {}
+      hash[:type] = 'ScheduleRulesetAbridged'
+      # set hash values from OpenStudio Object
+      hash[:identifier] = schedule_ruleset.nameString
```
chriswmackey, line 43: Since the original PR, I [added a method to escape Non-ASCII characters](https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/230), which should probably be used here to avoid situations of invalid JSONs.

lib/from_openstudio/schedule/ruleset.rb
```diff
@@ -0,0 +1,114 @@
...
+      hash[:default_day_schedule] = schedule_ruleset.defaultDaySchedule.nameString 
+      hash[:summer_designday_schedule] = schedule_ruleset.summerDesignDaySchedule.nameString
+      hash[:winter_designday_schedule] = schedule_ruleset.winterDesignDaySchedule.nameString
+      hash[:holiday_schedule] = schedule_ruleset.holidaySchedule.nameString
+
+      schedule_days = [schedule_ruleset.defaultDaySchedule, schedule_ruleset.summerDesignDaySchedule, schedule_ruleset.winterDesignDaySchedule, schedule_ruleset.holidaySchedule]
+      hash[:day_schedules] = []
+      schedule_days.each do |schedule_day|
+        hash[:day_schedules] << ScheduleRulesetAbridged.from_day_schedule(schedule_day)
+      end
```
chriswmackey, line 54: We also need to add all of the ScheduleDays that are used by each rule. And we should probably check for the case that the same ScheduleDay is used multiple times. I think this only needs to be a check for the object nameString and we can keep a set of day schedule names to check against.

lib/from_openstudio/schedule/type_limit.rb
```diff
@@ -0,0 +1,63 @@
...
+        hash[:lower_limit] = schedule_type_limit.lowerLimitValue.get
+      end
+      # check if boost optional object is empty
+      unless schedule_type_limit.upperLimitValue.empty?
+        hash[:upper_limit] = schedule_type_limit.upperLimitValue.get
+      end
+      # check if boost optional object is empty
+      unless schedule_type_limit.numericType.empty?
+        numeric_type = schedule_type_limit.numericType.get
+        hash[:numeric_type] = numeric_type.titleize
```
chriswmackey, line 55: Nice! That `.titleize` method seems like the right way to handle these enumeration variables when only the first letter needs to be capitalized.

lib/from_openstudio/schedule/type_limit.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+      hash[:unit_type] = schedule_type_limit.unitType
```
chriswmackey, line 57: We should also `.titleize` this one and then do a check for the few cases where the [enumeration variabe](https://www.ladybug.tools/honeybee-schema/model.html#tag/scheduletypelimit_model) has two words like `ActivityLevel`.

chriswmackey (approved):
Hey @tanushree04 ,

Thanks for putting this one together and sorry that it took me a while to review. You see that I have a few comments but they're relatively minor and merging the PR shouldn't break anything.

So I'm just going to merge it and I'll send another PR shortly that addresses the few tweaks I mentioned in my comments. Thanks for all of your help here!

PR #245, opened 2021-10-12
feat(displayname): Adds display names to objects

### LINKED ISSUE #89: Transfer display_name to OpenStudio objects
All of the objects that have an `identifier` in the honeybee Model also have a `display_name`.  We should try to add these display names to the OpenStudio objects.

### LINKED ISSUE #234: Make sure to titleize the Material roughness enumeration

tanushree04:
Resolves #89 
Resolves #234

### DIFF
--- a/lib/honeybee/model_object.rb
+++ b/lib/honeybee/model_object.rb
@@ -43,6 +43,7 @@ module Honeybee
       :universal_newline => true       # Always break lines with \n
     }
 
+
     def method_missing(sym, *args)
       name = sym.to_s
       aname = name.sub('=', '')
@@ -116,6 +117,17 @@ module Honeybee
       encode_str.gsub(/[^.A-Za-z0-9_-]/, '_').gsub(' ', '_')
     end
 
+    # Create methods to get and set display name 
+    if !OpenStudio::Model::ModelObject.method_defined?(:displayName)
+      OpenStudio::Model::ModelObject.class_eval do
+        define_method(:displayName) do
+          additionalProperties.getFeatureAsString("DisplayName")
+        end
+        define_method(:setDisplayName) do |value|
+          additionalProperties.setFeature("DisplayName", value)
+        end
+      end
+    end
 
   end # ModelObject
 end # Honeybee
--- a/lib/honeybee/model_object.rb
+++ b/lib/honeybee/model_object.rb
@@ -65,7 +65,8 @@ module Honeybee
     def self.read_from_disk(file)
       hash = nil
       File.open(File.join(file), 'r') do |f|
-        hash = JSON.parse(f.read, symbolize_names: true)
+        hash = JSON.parse(File.read(f, :external_encoding => 'iso-8859-1',
+          :internal_encoding => 'utf-8'), symbolize_names: true, encoding: 'UTF-8')
       end
       new(hash)
     end
--- a/lib/to_openstudio/construction/air.rb
+++ b/lib/to_openstudio/construction/air.rb
@@ -46,6 +46,9 @@ module Honeybee
       os_construction = OpenStudio::Model::ConstructionAirBoundary.new(openstudio_model)
       os_construction.setName(@hash[:identifier])
       os_construction.setAirExchangeMethod('None')
+      unless @hash[:display_name].nil?
+        os_construction.setDisplayName(@hash[:display_name])
+      end
 
       os_construction
     end
--- a/lib/to_openstudio/geometry/aperture.rb
+++ b/lib/to_openstudio/geometry/aperture.rb
@@ -114,6 +114,10 @@ module Honeybee
           os_subsurface.setName(@hash[:identifier])
         end
 
+        unless @hash[:display_name].nil?
+          os_subsurface.setDisplayName(@hash[:display_name])
+        end
+
         # assign the construction if it exists
         if @hash[:properties][:energy][:construction]
           construction_identifier = @hash[:properties][:energy][:construction]
--- a/spec/samples/program_type/program_type_abridged_kitchen.json
+++ b/spec/samples/program_type/program_type_abridged_kitchen.json
@@ -1,9 +1,11 @@
 {
     "type": "ProgramTypeAbridged",
     "identifier": "2013::FullServiceRestaurant::Kitchen",
+    "display_name": "~2013::FullServiceRestaurant::Kitchen",
     "people": {
         "type": "PeopleAbridged",
         "identifier": "2013::FullServiceRestaurant::Kitchen_People",
+        "display_name": "//2013::FullServiceRestaurant::Kitchen_People",
         "people_per_area": 0.05381957525591208,
         "radiant_fraction": 0.3,
         "latent_fraction": {
@@ -15,6 +17,7 @@
     "lighting": {
         "type": "LightingAbridged",
         "identifier": "2013::FullServiceRestaurant::Kitchen_Lighting",
+        "display_name": "//2013::FullServiceRestaurant::Kitchen_Lighting//",
         "watts_per_area": 13.024318999999998,
         "return_air_fraction": 0.0,
         "radiant_fraction": 0.7,
--- a/spec/tests/honeybee_hvac_spec.rb
+++ b/spec/tests/honeybee_hvac_spec.rb
@@ -49,6 +49,8 @@ RSpec.describe Honeybee do
     object1 = honeybee_obj_1.to_openstudio(openstudio_model)
     expect(object1).not_to be nil
     expect(object1.nameString).to eq 'Passive House HVAC System'
+    expect((object1.additionalProperties.featureNames)).to include ("DisplayName")
+    #expect((object1.displayName.get)).to eq 'Test Name áéíóúüñ¿¡'
     expect(object1.maximumHeatingSupplyAirTemperature).to eq 40.0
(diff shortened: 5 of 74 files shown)


lib/to_openstudio/geometry/aperture.rb
```diff
@@ -114,6 +114,11 @@ def to_openstudio(openstudio_model)
           os_subsurface.setName(@hash[:identifier])
         end
 
+        #TODO: check
+        unless @hash[:display_name].nil?
+          os_subsurface.additionalProperties().setFeature("Display Name", @hash[:display_name])
+        end
```
tanushree04, line 120: Check

spec/samples/program_type/program_type_abridged_kitchen.json
```diff
@@ -1,9 +1,11 @@
 {
     "type": "ProgramTypeAbridged",
     "identifier": "2013::FullServiceRestaurant::Kitchen",
+    "display_name": "~2013::FullServiceRestaurant::Kitchen",
     "people": {
         "type": "PeopleAbridged",
         "identifier": "2013::FullServiceRestaurant::Kitchen_People",
+        "display_name": "//2013::FullServiceRestaurant::Kitchen_People",
```
macumber, line 8: You might make some tests with Unicode characters, commas, or other unusual characteristics to test that serialization works correctly

spec/tests/honeybee_model_spec.rb
```diff
@@ -76,6 +76,7 @@
     expect(openstudio_space.empty?).to be false
     openstudio_space = openstudio_space.get
     expect(openstudio_space.nameString).to eq 'Tiny_House_Office'
+    expect((openstudio_space.additionalProperties.featureNames)).to include ("Display Name")
```
macumber, line 79: Might want to check the value is correct too

lib/to_openstudio/construction/air.rb
```diff
@@ -45,6 +45,9 @@ def find_existing_openstudio_object(openstudio_model)
     def to_openstudio(openstudio_model)
       os_construction = OpenStudio::Model::ConstructionAirBoundary.new(openstudio_model)
       os_construction.setName(@hash[:identifier])
+      unless @hash[:display_name].nil?
+        os_construction.additionalProperties().setFeature("Display Name", @hash[:display_name])
```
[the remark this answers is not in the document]
tanushree04: @macumber This is added as a temporary fix till there is a new method. That makes sense, I can transfer this to the ModelObject, that way once there is a new method it would only need to be updated there.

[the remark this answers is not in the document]
macumber: Cool.  Just to be clear, the model object method would still store data in the additionalProperties like you are now.  It would mostly just hard code the key name in C++

One other question, are you sure you want a space in the key name?  E.g. “DisplayName” vs “Display Name”?

spec/tests/honeybee_hvac_spec.rb
```diff
@@ -49,6 +49,8 @@
     object1 = honeybee_obj_1.to_openstudio(openstudio_model)
     expect(object1).not_to be nil
     expect(object1.nameString).to eq 'Passive House HVAC System'
+    expect((object1.additionalProperties.featureNames)).to include ("DisplayName")
+    #expect((object1.displayName.get)).to eq 'Test Name áéíóúüñ¿¡'
```
tanushree04, line 53: The commented out tests with non English characters are failing I think there is an issue with JSON to Hash serialization. On printing out the encoding my ruby environment is using UTF-8.

spec/tests/osm_construction_set_spec.rb
```diff
@@ -51,6 +53,7 @@
     expect(honeybee).not_to be nil
     expect(honeybee[0][:type]).to eq 'ConstructionSetAbridged'
     expect(honeybee[0][:identifier]).to eq '189.1-2009 - CZ1 - Office'
+    #expect(honeybee[0][:display_name]).to eq 'Test Name áéíóúüñ¿¡'
```
tanushree04, line 56: @chriswmackey I tried to use the to_json instead of the pretty json method but I am still getting an error

     Failure/Error: expect(honeybee[0][:display_name]).to eq 'Test Name ├í├⌐├¡├│├║├╝├▒┬┐┬í'

       expected: "Test Name \u00E1\u00E9\u00ED\u00F3\u00FA\u00FC\u00F1\u00BF\u00A1"
            got: "Test Name \xC3\xA1\xC3\xA9\xC3\xAD\xC3\xB3\xC3\xBA\xC3\xBC\xC3\xB1\xC2\xBF\xC2\xA1"

chriswmackey: Hey, @tanushree04 ,

Sorry for the late response. If you aren't getting the same error that I did, at least we can say that the JSON is able to be written when using the `to_json` method.

It looks like the `to_json` method is escaping the characters as hex codes instead of unicodes. [See here for a reference](https://stackoverflow.com/questions/33415388/whats-the-difference-between-hex-code-x-and-unicode-u-chars).

I just checked and the Python JSON modules that we are using really need the characters to be escaped as unicode (using the hex code results in an error).

@macumber , could you recommend a way to ensure that the characters are escaped as unicode characters and not hex characters?

tanushree04: @chriswmackey (construction_set.displayName.get).force_encoding("UTF-8") seems to work, I will push the changes.

spec/tests/honeybee_hvac_spec.rb
tanushree04: @macumber do you know how I can retain non-ascii characters when going from honeybee JSON to OpenStudio?
When translating from OpenStudio to JSON - I was able to use this `(construction_set.displayName.get).force_encoding("UTF-8")`. This doesn't however seem to work the other way round for example `os_constr_set.setDisplayName(@hash[:display_name].force_encoding("UTF-8"))`. 

This is the error I see: 

Failure/Error: expect((object1.displayName.get)).to eq 'Ω▒┤∞äñδ¬à'

       expected: "\uAC74\uC124\uBA85"
            got: "\xCE\xA9\xE2\x96\x92\xE2\x94\xA4\xE2\x88\x9E\xC3\xA4\xC3\xB1\xCE\xB4\xC2\xAC\xC3\xA0"

       (compared using ==)

tanushree04: I think the error is in retaining non-ascii characters in a ruby hash, because when I try to set a string with special characters it is set as the display name correctly

lib/honeybee/model_object.rb
```diff
@@ -64,7 +65,8 @@ def method_missing(sym, *args)
     def self.read_from_disk(file)
       hash = nil
       File.open(File.join(file), 'r') do |f|
-        hash = JSON.parse(f.read, symbolize_names: true)
+        hash = JSON.parse(File.read(f, :external_encoding => 'iso-8859-1',
+          :internal_encoding => 'utf-8'), symbolize_names: true, encoding: 'UTF-8')
```
tanushree04, line 69: This helped - when I print this hash I see Unicode characters. Somehow this is not translating to the object classes hash (https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/245/files#diff-2296abac85df1839aba12ad4a880f93c287d91392eacc4c21cbfea9cfe9c5949R53)

spec/tests/honeybee_hvac_spec.rb
macumber: Hey @tanushree04 did you figure this out?  If not, I would guess that the issue is needing to set the encoding on the string to UTF8 before putting it into the honeybee hash or something like that?

lib/honeybee/model_object.rb
macumber: This looks like you are on the right track.  Is this reproducible on a local WIndows machine or does it only fail on CI?

tanushree04: @macumber you can reproduce it by uncommenting this test: https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/245/files#diff-2296abac85df1839aba12ad4a880f93c287d91392eacc4c21cbfea9cfe9c5949R53

macumber: Try this out: https://github.com/tanushree04/honeybee-openstudio-gem/pull/1

tanushree04: @macumber this worked for me. Thanks for all your help!

tanushree04: I merged your PR to this.

lib/honeybee/model_object.rb
```diff
@@ -116,6 +132,21 @@ def self.clean_identifier(str)
       encode_str.gsub(/[^.A-Za-z0-9_-]/, '_').gsub(' ', '_')
     end
 
+    # Create methods to get and set display name
+    if !OpenStudio::Model::ModelObject.method_defined?(:displayName)
```
chriswmackey, line 136: I just wanted to say that I really like the way this is implemented. It's very clean and will be much easier to maintain over time!

chriswmackey (approved):
This looks great, @tanushree04 . You've added a lot of value here that many users will appreciate.

Thank you, also, @macumber, for your reviews and suggestions.

Let's merge!

macumber:
Nice, congrats!

PR #251, opened 2021-11-29
feat(ProgramType): Adds Reverse Translators for ProgramType and Loads

### LINKED ISSUE #178: Add translators for SpaceTypes, Zone Infiltration, Ventilation and Setpoints to HBJSON ProgramTypes
This can only be completed after this issue is addressed:
https://github.com/ladybug-tools/honeybee-openstudio-gem/issues/177

This essentially includes a reverse translator for this module:
https://github.com/ladybug-tools/honeybee-openstudio-gem/blob/master/lib/to_openstudio/program_type.rb
... though I also imagine that we might be doing a lot of these edits through a reverse translator for the Room model, as can be seen in the forward translator:
https://github.com/ladybug-tools/honeybee-openstudio-gem/blob/master/lib/to_openstudio/geometry/room.rb

We also need tests.

tanushree04:
Resolves #178

### DIFF
--- a/lib/from_openstudio/geometry/room.rb
+++ b/lib/from_openstudio/geometry/room.rb
@@ -64,12 +64,145 @@ module Honeybee
       hash = {}
       hash[:type] = 'RoomPropertiesAbridged'
       hash[:energy] = self.energy_properties_from_space(space)
+
       hash
     end
 
     def self.energy_properties_from_space(space)
       hash = {}
       hash[:type] = 'RoomEnergyPropertiesAbridged'
+      # set room energy properties
+      unless space.defaultConstructionSet.empty?
+        const_set = space.defaultConstructionSet.get
+        hash[:construction_set] = const_set.nameString
+      end
+      unless space.spaceType.empty?
+        space_type = space.spaceType.get
+        hash[:program_type] = space_type.nameString
+      end
+      # TODO: These are loads assigned to the space directly. How should duplicates created in programtype, if any, be handled? 
+      unless space.people.empty?
+        space.people.each do |people|
+          # Only translate if people per floor area is specified
+          unless people.peopleDefinition.peopleperSpaceFloorArea.empty?
+            hash[:people] = Honeybee::PeopleAbridged.from_load(people)
+            break
+          end
+        end
+      end
+      unless space.lights.empty?
+        space.lights.each do |light|
+          # Only translate if watts per floor area is specified
+          unless light.lightsDefinition.wattsperSpaceFloorArea.empty?
+            hash[:lighting] = Honeybee::LightingAbridged.from_load(light)
+            break
+          end
+        end
+      end
+      unless space.electricEquipment.empty?
+        space.electricEquipment.each do |electric_eq|
+          electric_eq_def = electric_eq.electricEquipmentDefinition
+          # Only translate if watts per floor area is specified
+          unless electric_eq_def.wattsperSpaceFloorArea.empty?
+            hash[:electric_equipment] = Honeybee::ElectricEquipmentAbridged.from_load(electric_eq)
+            break
+          end
+        end
+      end
+      unless space.gasEquipment.empty?
+        space.gasEquipment.each do |gas_eq|
+          gas_eq_def = gas_eq.gasEquipmentDefinition
+          unless gas_eq_def.wattsperSpaceFloorArea
+            hash[:gas_equipment] = Honeybee::GasEquipmentAbridged.from_load(gas_eq)
+            break
+          end
+        end
+      end
+      unless space.otherEquipment.empty?
+        hash[:process_loads] = []
+        space.otherEquipment.each do |other_eq|
+          unless other_eq.designLevel.empty?
+            hash[:process_loads] << Honeybee::ProcessAbridged.from_load(other_eq)
+          end
+        end
+      end
+      unless space.spaceInfiltrationDesignFlowRates.empty?
+        space.spaceInfiltrationDesignFlowRates.each do |infilt|
+          # Only translate if flow per exterior area is specified
+          unless infilt.flowperExteriorSurfaceArea.empty?
+            hash[:infiltration] = Honeybee::InfiltrationAbridged.from_load(infilt)
+            break
+          end
+        end
+      end
+      unless space.designSpecificationOutdoorAir.empty?
+        hash[:ventilation] = Honeybee::VentilationAbridged.from_load(space.designSpecificationOutdoorAir.get)
+      end
+      unless space.daylightingControls.empty?
+        hash[:daylighting_control] = Honeybee::DaylightingControl.from_load(space.daylightingControls[0])
+      end
+      thermal_zone = space.thermalZone
+      unless thermal_zone.empty?
+        thermal_zone = space.thermalZone.get
+        unless thermal_zone.thermostatSetpointDualSetpoint.empty?
+        # TODO: There isn't a combined setpoint object in OS and the identifier can't be assigned.
+        # For now using the thermal zone name as setpoint identifier.
+          hash[:setpoint] = {}
+          thermostat = thermal_zone.thermostatSetpointDualSetpoint.get
+          hash[:setpoint][:identifier] = thermostat.nameString
+          if thermostat.heatingSetpointTemperatureSchedule.empty?
+            # if heating setpoint schedule is not specified create a new setpoint schedule and assign to HB thermostat object.
+            # first check if schedule is already created
+            if $heating_setpoint_schedule.nil?
+              openstudio_model = OpenStudio::Model::Model.new
+              openstudio_schedule = OpenStudio::Model::ScheduleRuleset.new(openstudio_model)
+              openstudio_schedule.setName('Heating Schedule Default')
+              openstudio_sch_type_lim = OpenStudio::Model::ScheduleTypeLimits.new(openstudio_model)
+              openstudio_sch_type_lim.setName('Temperature')
+              openstudio_sch_type_lim.setNumericType('Temperature')
+              openstudio_schedule.defaultDaySchedule.setName('Heating Day Default')
+              openstudio_schedule.defaultDaySchedule.addValue(OpenStudio::Time.new(0,24,0,0), 100)
+              openstudio_schedule.defaultDaySchedule.setScheduleTypeLimits(openstudio_sch_type_lim)
+              $heating_setpoint_schedule = Honeybee::ScheduleRulesetAbridged.from_schedule_ruleset(openstudio_schedule)
+            end
+            hash[:setpoint][:heating_schedule] = $heating_setpoint_schedule[:identifier]
+          else
+            heating_schedule = thermostat.heatingSetpointTemperatureSchedule.get
+            hash[:setpoint][:heating_schedule] = heating_schedule.nameString
+          end
+          if thermostat.coolingSetpointTemperatureSchedule.empty?
+            # if cooling setpoint schedule is not specified create a new setpoint schedule and assign to HB thermostat object
+            # first check if schedule is already created
+            if $cooling_setpoint_schedule.nil?
+              openstudio_model = OpenStudio::Model::Model.new
+              openstudio_schedule = OpenStudio::Model::ScheduleRuleset.new(openstudio_model)
+              openstudio_schedule.setName('Cooling Schedule Default')
+              openstudio_sch_type_lim = OpenStudio::Model::ScheduleTypeLimits.new(openstudio_model)
+              openstudio_sch_type_lim.setName('Temperature')
+              openstudio_sch_type_lim.setNumericType('Temperature')
+              openstudio_schedule.defaultDaySchedule.setName('Cooling Day Default')
+              openstudio_schedule.defaultDaySchedule.addValue(OpenStudio::Time.new(0,24,0,0), -100)
+              openstudio_schedule.defaultDaySchedule.setScheduleTypeLimits(openstudio_sch_type_lim)
+              $cooling_setpoint_schedule = Honeybee::ScheduleRulesetAbridged.from_schedule_ruleset(openstudio_schedule)
+            end
+            hash[:setpoint][:cooling_schedule] = $cooling_setpoint_schedule[:identifier]
+          else
+            cooling_schedule = thermostat.coolingSetpointTemperatureSchedule.get
+            hash[:setpoint][:cooling_schedule] = cooling_schedule.nameString
+          end
+        end
+        unless thermal_zone.zoneControlHumidistat.empty?
+          humidistat = thermal_zone.zoneControlHumidistat.get
+          unless humidistat.humidifyingRelativeHumiditySetpointSchedule.empty?
+            humidifying_schedule = humidistat.humidifyingRelativeHumiditySetpointSchedule.get
+            hash[:setpoint][:humidifying_schedule] = humidifying_schedule.nameString
+          end
+          unless humidistat.dehumidifyingRelativeHumiditySetpointSchedule.empty?
+            dehumidifying_schedule = humidistat.dehumidifyingRelativeHumiditySetpointSchedule.get
+            hash[:setpoint][:dehumidifying_schedule] = dehumidifying_schedule.nameString
+          end
+        end
+      end 
       hash
     end
 
--- a/lib/from_openstudio/geometry/room.rb
+++ b/lib/from_openstudio/geometry/room.rb
@@ -83,19 +83,29 @@ module Honeybee
       # TODO: These are loads assigned to the space directly. How should duplicates created in programtype, if any, be handled? 
       unless space.people.empty?
         space.people.each do |people|
+          people_def = people.peopleDefinition
           # Only translate if people per floor area is specified
-          unless people.peopleDefinition.peopleperSpaceFloorArea.empty?
-            hash[:people] = Honeybee::PeopleAbridged.from_load(people)
-            break
+          # Check if schedule exists and is of the correct type
+          if !people_def.peopleperSpaceFloorArea.empty? && !people.numberofPeopleSchedule.empty?
+            sch = people.numberofPeopleSchedule.get
+            if sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized
+              hash[:people] = Honeybee::PeopleAbridged.from_load(people)
+              break
+            end
           end
         end
       end
       unless space.lights.empty?
         space.lights.each do |light|
+          light_def = light.lightsDefinition
           # Only translate if watts per floor area is specified
-          unless light.lightsDefinition.wattsperSpaceFloorArea.empty?
-            hash[:lighting] = Honeybee::LightingAbridged.from_load(light)
-            break
+          # Check if schedule exists and is of the correct type
+          if !light_def.wattsperSpaceFloorArea.empty? && !light.schedule.empty?
+            sch = light.schedule.get
+            if sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized
+              hash[:lighting] = Honeybee::LightingAbridged.from_load(light)
+              break
+            end
           end
         end
       end
@@ -103,18 +113,27 @@ module Honeybee
         space.electricEquipment.each do |electric_eq|
           electric_eq_def = electric_eq.electricEquipmentDefinition
           # Only translate if watts per floor area is specified
-          unless electric_eq_def.wattsperSpaceFloorArea.empty?
-            hash[:electric_equipment] = Honeybee::ElectricEquipmentAbridged.from_load(electric_eq)
-            break
+          # Check if schedule exists and is of the correct type
+          if !electric_eq_def.wattsperSpaceFloorArea.empty? && !electric_eq.schedule.empty?
+            sch = electric_eq.schedule.get
+            if sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized
+              hash[:electric_equipment] = Honeybee::ElectricEquipmentAbridged.from_load(electric_eq)
+              break
+            end
           end
         end
       end
       unless space.gasEquipment.empty?
         space.gasEquipment.each do |gas_eq|
           gas_eq_def = gas_eq.gasEquipmentDefinition
-          unless gas_eq_def.wattsperSpaceFloorArea
-            hash[:gas_equipment] = Honeybee::GasEquipmentAbridged.from_load(gas_eq)
-            break
+          # Only translate if watts per floor area is specified
+          # Check if schedule exists and is of the correct type
+          if !gas_eq_def.wattsperSpaceFloorArea.empty? && !gas_eq.schedule.empty?
+            sch = gas_eq.schedule.get
+            if sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized
+              hash[:gas_equipment] = Honeybee::GasEquipmentAbridged.from_load(gas_eq)
+              break
+            end
           end
         end
       end
@@ -127,11 +146,15 @@ module Honeybee
         end
       end
       unless space.spaceInfiltrationDesignFlowRates.empty?
-        space.spaceInfiltrationDesignFlowRates.each do |infilt|
+        space.spaceInfiltrationDesignFlowRates.each do |infiltration|
           # Only translate if flow per exterior area is specified
-          unless infilt.flowperExteriorSurfaceArea.empty?
-            hash[:infiltration] = Honeybee::InfiltrationAbridged.from_load(infilt)
-            break
+          # Check if schedule exists and is of the correct type
+          if !infiltration.flowperExteriorSurfaceArea.empty? && !infiltration.schedule.empty?
+            sch = infiltration.schedule.get
+            if sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized
+              hash[:infiltration] = Honeybee::InfiltrationAbridged.from_load(infiltration)
+              break
+            end
           end
         end
       end
@@ -145,8 +168,6 @@ module Honeybee
       unless thermal_zone.empty?
         thermal_zone = space.thermalZone.get
         unless thermal_zone.thermostatSetpointDualSetpoint.empty?
-        # TODO: There isn't a combined setpoint object in OS and the identifier can't be assigned.
-        # For now using the thermal zone name as setpoint identifier.
           hash[:setpoint] = {}
           thermostat = thermal_zone.thermostatSetpointDualSetpoint.get
           hash[:setpoint][:identifier] = thermostat.nameString
--- a/lib/from_openstudio/load/electric_equipment.rb
+++ b/lib/from_openstudio/load/electric_equipment.rb
@@ -29,56 +29,36 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *******************************************************************************
 
-# import the honeybee objects which we will extend
-require 'honeybee'
+require 'honeybee/load/electric_equipment'
+require 'to_openstudio/model_object'
 
-# extend the compound objects that house the other objects
-require 'from_openstudio/model'
-require 'from_openstudio/model_object'
-require 'from_openstudio/construction_set'
+module Honeybee
+    class ElectricEquipmentAbridged < ModelObject
 
-# extend the geometry objects
-require 'from_openstudio/geometry/aperture'
-require 'from_openstudio/geometry/door'
-require 'from_openstudio/geometry/face'
-require 'from_openstudio/geometry/room'
-require 'from_openstudio/geometry/shade'
+        def self.from_load(load)
+            # create an empty hash
+            hash = {}
+            hash[:type] = 'ElectricEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                hash[:schedule] = schedule.name.to_s
+            end
+            load_def = load.electricEquipmentDefinition
+            unless load_def.wattsperSpaceFloorArea.empty?
+                hash[:watts_per_area] = load_def.wattsperSpaceFloorArea.get
+            end
+            hash[:radiant_fraction] = load_def.fractionRadiant
+            hash[:latent_fraction] = load_def.fractionLatent
+            hash[:lost_fraction] = load_def.fractionLost
 
-# extend the construction objects
-require 'from_openstudio/construction/opaque'
-require 'from_openstudio/construction/window'
-require 'from_openstudio/construction/shade'
-require 'from_openstudio/construction/air'
+            hash
+        end
 
-# import the material objects
-require 'from_openstudio/material/opaque'
-require 'from_openstudio/material/opaque_no_mass'
-require 'from_openstudio/material/window_gas'
-require 'from_openstudio/material/window_gas_mixture'
-require 'from_openstudio/material/window_gas_custom'
-require 'from_openstudio/material/window_blind'
-require 'from_openstudio/material/window_glazing'
-require 'from_openstudio/material/window_simpleglazsys'
 
-# extend the simulation objects
-require 'from_openstudio/simulation/design_day'
-require 'from_openstudio/simulation/parameter_model'
-require 'from_openstudio/simulation/simulation_output'
-
-# extend the schedule objects
-require 'from_openstudio/schedule/type_limit'
-require 'from_openstudio/schedule/ruleset'
-require 'from_openstudio/schedule/fixed_interval'
-
-# extend the load objects
-require 'from_openstudio/load/electric_equipment'
-require 'from_openstudio/load/people'
-require 'from_openstudio/load/gas_equipment'
-require 'from_openstudio/load/lighting'
-require 'from_openstudio/load/infiltration'
-require 'from_openstudio/load/ventilation'
-require 'from_openstudio/load/daylight'
-require 'from_openstudio/load/process'
-
-# extend the program type objects
-require 'from_openstudio/program_type'
+    end #ElectricEquipmentAbridged
+end #Honeybee
--- a/lib/from_openstudio/load/electric_equipment.rb
+++ b/lib/from_openstudio/load/electric_equipment.rb
@@ -46,7 +46,9 @@ module Honeybee
             end
             unless load.schedule.empty?
                 schedule = load.schedule.get
-                hash[:schedule] = schedule.name.to_s
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
+                    hash[:schedule] = schedule.nameString
+                end
             end
             load_def = load.electricEquipmentDefinition
             unless load_def.wattsperSpaceFloorArea.empty?
--- a/lib/from_openstudio/load/gas_equipment.rb
+++ b/lib/from_openstudio/load/gas_equipment.rb
@@ -29,56 +29,32 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *******************************************************************************
 
-# import the honeybee objects which we will extend
-require 'honeybee'
+require 'honeybee/load/gas_equipment'
+require 'to_openstudio/model_object'
 
-# extend the compound objects that house the other objects
-require 'from_openstudio/model'
-require 'from_openstudio/model_object'
-require 'from_openstudio/construction_set'
+module Honeybee
+    class GasEquipmentAbridged
 
-# extend the geometry objects
-require 'from_openstudio/geometry/aperture'
-require 'from_openstudio/geometry/door'
-require 'from_openstudio/geometry/face'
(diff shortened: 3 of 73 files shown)


lib/from_openstudio/geometry/room.rb
```diff
@@ -70,6 +70,82 @@ def self.properties_from_space(space)
...
+      unless thermal_zone.empty?
+        thermal_zone = space.thermalZone.get
+        # TODO: The schema for setpoint looks incorrect - there isn't a combined setpoint object in OS. The identifier can't be assigned.
+        # For now using the thermal zone name as setpoint identifier.
+        hash[:setpoint] = {}
+        hash[:setpoint][:identifier] = thermal_zone.nameString
+        unless thermal_zone.thermostatSetpointDualSetpoint.empty?
+          thermostat = thermal_zone.thermostatSetpointDualSetpoint.get
+          #TODO: heating_schedule/cooling_schedule is an optional property in OS Thermostat object. This should not be a required property in the schema.
+          unless thermostat.heatingSetpointTemperatureSchedule.empty?
```
tanushree04, line 128: @chriswmackey In the OS SDK Heating and Cooling schedule are optional properties for the [thermostat object](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.3.0-doc/model/html/classopenstudio_1_1model_1_1_thermostat_setpoint_dual_setpoint.html). These are however required properties. 
This will throw a schema validation error for the reverse translators if these aren't set in the OS model.
Any thoughts on how to avoid that?

lib/from_openstudio/load/electric_equipment.rb
```diff
@@ -0,0 +1,65 @@
...
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                hash[:schedule] = schedule.name.to_s
+            end
+            load_def = load.electricEquipmentDefinition
+            #TODO: This is an optional property in OS but a required property in the schema. It will through an error if not specified in OS. 
+            unless load_def.wattsperSpaceFloorArea.empty?
```
tanushree04, line 53: @chriswmackey similarly here, [watts per area](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.3.0-doc/model/html/classopenstudio_1_1model_1_1_electric_equipment_definition.html) is an optional property in the OS SDK but is required by our schema.

Is there a default that we can set here to avoid a schema validation error?

lib/from_openstudio/program_type.rb
```diff
@@ -0,0 +1,75 @@
...
+            #unless programtype.hotWaterEquipment.empty?
+            #    hash[:service_hot_water] = Honeybee::ServiceHotWaterAbridged.from_load(programtype.hotWaterEquipment[0])
+            #end
+            unless programtype.spaceInfiltrationDesignFlowRates.empty?
+                hash[:infiltration] = Honeybee::InfiltrationAbridged.from_load(programtype.spaceInfiltrationDesignFlowRates[0])
+            end
+            unless programtype.designSpecificationOutdoorAir.empty?
+                hash[:ventilation] = Honeybee::VentilationAbridged.from_load(programtype.designSpecificationOutdoorAir.get)
+            end
+            # TODO: add setpoint thermostat and humidistat? Not sure if a reverse translator would have a programtype setpoint object - These are assigned at the thermal zone level.
```
tanushree04, line 70: @chriswmackey Do you have thoughts on this? These objects are only defined at the thermal zone level in the OS SDK and we are translating those to Honeybee rooms. How do we want to handle adding these objects to the programtype?

spec/tests/osm_material_spec.rb
```diff
@@ -36,6 +36,12 @@ def load_file(file_name)
   file = File.join(File.dirname(__FILE__), '../samples/osm_material/', file_name)
   vt = OpenStudio::OSVersion::VersionTranslator.new
   osm = vt.loadModel(file)
+  year = 2020
+  osm.get.getYearDescription.setCalendarYear(year.to_i)
+  weather_file = File.join(File.dirname(__FILE__), '../samples/epw')
+  workflow = OpenStudio::WorkflowJSON.new
+  workflow.setSeedFile(file)
+  workflow.setWeatherFile(File.absolute_path(weather_file))
```
tanushree04, line 44: @joseph-robertson I am setting the calendar year and also setting the weather file here. I am still seeing the error: 

`[openstudio.model.YearDescription] <1> 'UseWeatherFile' is selected in YearDescription, but there are no 
weather file set for the model.`

From the code link you shared - this should not be the case. Do you have any thoughts on what could be happening?

lib/from_openstudio/geometry/room.rb
chriswmackey: Hm. I see that EnergyPlus also interprets the Thermostat as valid/simulate-able when there are no setpoints on the thermostat object, which seems odd. If a given schedule is missing, can we just create a dummy setpoint schedule with extreme values. For example 100C for the cooling setpoint and -100C for the heating setpoint?

lib/from_openstudio/load/electric_equipment.rb
chriswmackey: The same goes for infiltration and the "per exterior area" specification. For Ventilation, we allow for the sum of all of the different types of specification. So we can import Ventilation no matter what specification method is used.

lib/from_openstudio/program_type.rb
chriswmackey: I think it's fine if Ventilation and Setpoint objects are not assign-able to ProgramTypes when importing from OSM/IDF. Let's just assign it to the Rooms and that will ensure that the resulting simulations are the same between the imported IDF and the existing IDF.

lib/from_openstudio/load/electric_equipment.rb
tanushree04: Okay, I have made these changes. 
Its also applicable to `watts_per_area` for lighting and `people_per_area` for people loads.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,129 @@ def self.properties_from_space(space)
...
+        end
+      end
+      thermal_zone = space.thermalZone
+      unless thermal_zone.empty?
+        thermal_zone = space.thermalZone.get
+        unless thermal_zone.thermostatSetpointDualSetpoint.empty?
+        # TODO: There isn't a combined setpoint object in OS and the identifier can't be assigned.
+        # For now using the thermal zone name as setpoint identifier.
+          hash[:setpoint] = {}
+          hash[:setpoint][:identifier] = thermal_zone.nameString
```
tanushree04, line 141: @chriswmackey I have used the thermal zone namestring as an identifier for setpoint

tanushree04:
@chriswmackey I have implemented the changes and this should be ready to review. 
The only pending item for spacetypes should be reverse translators for ServiceHotWater but please feel free to review the rest.

lib/from_openstudio/load/electric_equipment.rb
```diff
@@ -0,0 +1,64 @@
...
+            hash = {}
+            hash[:type] = 'ElectricEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                hash[:schedule] = schedule.name.to_s
```
chriswmackey, line 49: The schedule is a required attribute in honeybee-schema. So, if the schedule is ,empty?, then the resulting hash is invalid. I think we have two options for the case that there's no schedule:
1. Create a dummy "Always On" schedule and use that. I think we would implement this similar to how the `$heating_setpoint_schedule` should be implemented, which a check for .nil? before we try to create it.
2. Just perform a check for whether the schedule exists before we create the hash. If the schedule does not exist, then we just disregard the load. The check like this will be similar to the `unless electric_eq_def.wattsperSpaceFloorArea.empty?` that we currently perform before creating the load object.

I leave the decision up to you, @tanushree04 , about which one is easier to implement for you. But this current situation is going to produce hashes that are not serializable by the Honeybee Python library.

lib/from_openstudio/load/gas_equipment.rb
```diff
@@ -0,0 +1,60 @@
...
+            # create an empty hash
+            hash = {}
+            hash[:type] = 'GasEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.schedule.empty?
+                hash[:schedule] = (load.schedule.get).to_s
+            end
+            load_def = load.gasEquipmentDefinition
+            load_def.setWattsperSpaceFloorArea(15)
```
chriswmackey, line 48: What is going on here? Is this class just not finished yet? This whole class should be like the `ElectricEquipmentAbridged` class.

lib/from_openstudio/load/gas_equipment.rb
```diff
@@ -0,0 +1,60 @@
...
+    class GasEquipmentAbridged
+
+        def self.from_load(load)
+            # create an empty hash
+            hash = {}
+            hash[:type] = 'GasEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.schedule.empty?
+                hash[:schedule] = (load.schedule.get).to_s
```
chriswmackey, line 45: We'll need a check for the schedule before we create the object.

lib/from_openstudio/load/infiltration.rb
```diff
@@ -0,0 +1,62 @@
...
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.flowperExteriorSurfaceArea.empty?
+                hash[:flow_per_exterior_area] = load.flowperExteriorSurfaceArea.get
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                hash[:schedule] = schedule.nameString
```
chriswmackey, line 52: We'll need a check for the schedule before we create the object.

lib/from_openstudio/load/electric_equipment.rb
chriswmackey: We also need to make sure that the schedule is of an acceptable type. In other words, it has to be a ScheduleRuleset or a ScheduleFixedInterval. If it's not, then don't have translators for the schedule to the Honeybee hash so we'll have to either not translate the whole load object or we use an Always On schedule.

lib/from_openstudio/load/lighting.rb
```diff
@@ -0,0 +1,64 @@
...
+            hash = {}
+            hash[:type] = 'LightingAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                hash[:schedule] = schedule.name.to_s
```
chriswmackey, line 49: We need a check for the schedule here.

lib/from_openstudio/load/people.rb
```diff
@@ -0,0 +1,64 @@
...
+            hash[:type] = 'PeopleAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            #TODO: Is there a default occupancy schedule since it is a required field in HB but not
+            #in OS
+            unless load.numberofPeopleSchedule.empty?
+                hash[:occupancy_schedule] = (load.numberofPeopleSchedule.get).name.to_s
```
chriswmackey, line 50: We need a check for the schedule here.

lib/from_openstudio/load/people.rb
```diff
@@ -0,0 +1,64 @@
...
+            end
+            unless load.activityLevelSchedule.empty?
+                hash[:activity_schedule] = (load.activityLevelSchedule.get).name.to_s
```
chriswmackey, line 53: So the `activity_schedule` is optional and it's ok for it to be unspecified here. But, if the schedule exists and it is not a ScheudleRuleset or ScheduleFixedInterval, this will create an invalid HBJSON Model. So we need to check the type of schedule here.

lib/from_openstudio/load/people.rb
```diff
@@ -0,0 +1,64 @@
...
+            end
+            load_def = load.peopleDefinition
+            unless load_def.peopleperSpaceFloorArea.empty?
+                hash[:people_per_area] = load_def.peopleperSpaceFloorArea.to_f
+            end
+            hash[:radiant_fraction] = load_def.fractionRadiant
+            hash[:latent_fraction] = 1 - load_def.fractionRadiant
```
chriswmackey, line 60: This is not how the `latent_fraction` is calculated. To do this correctly, you should check the `isSensibleHeatFractionAutocalculated`. If it is false, then get the `sensibleHeatFraction` and (if it's not empty) set the latent fraction to `1 - load_def.sensibleHeatFraction`.

lib/from_openstudio/load/process.rb
```diff
@@ -0,0 +1,60 @@
...
+            hash = {}
+            hash[:type] = 'ProcessAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            load_def = load.otherEquipmentDefinition
+            unless load_def.displayName.empty?
+                hash[:display_name] = (load_def.displayName.get).force_encoding("UTF-8")
+            end
+            hash[:watts] = load_def.getDesignLevel
+            hash[:schedule] = load.schedule.nameString
```
chriswmackey, line 49: Hm. Is the OtherEquipment.schedule property never optional? At the least we need a check to make sure the schedule if of a correct type.

lib/from_openstudio/load/ventilation.rb
```diff
@@ -0,0 +1,57 @@
...
+            hash[:type] = 'VentilationAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            hash[:air_changes_per_hour] = load.outdoorAirFlowAirChangesperHour
+            hash[:flow_per_zone] = load.outdoorAirFlowRate
+            hash[:flow_per_person] = load.outdoorAirFlowperPerson
+            hash[:flow_per_area] = load.outdoorAirFlowperFloorArea
+            unless load.outdoorAirFlowRateFractionSchedule.empty?
+                load = load.outdoorAirFlowRateFractionSchedule.get
+                hash[:schedule] = load.name
```
chriswmackey, line 50: This is mostly good here but we need a check to ensure that the schedule is of the correct type.

lib/from_openstudio/model.rb
```diff
@@ -65,11 +66,12 @@ def self.translate_from_openstudio(openstudio_model)
       hash[:tolerance] = 0.01
       hash[:angle_tolerance] = 1.0
 
-      # Hashes for all shcedules and constructions in the model
+      # Hashes for various model properties
       $schedules = {}
       $opaque_constructions = {}
       $window_constructions = {}
       $shade_constructions = {}
+      $programtype_setpoint_hash = {}
```
chriswmackey, line 74: In OpenStudio, thermostats are not assignable to space types (there's always going to be a separate thermostat for each zone). So this seems like a variable that is not needed.

lib/from_openstudio/program_type.rb
```diff
@@ -0,0 +1,91 @@
...
+        def self.from_programtype(programtype)
+            # create an empty hash
+            hash = {}
+            hash[:type] = 'ProgramTypeAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(programtype.nameString)
+            unless programtype.displayName.empty?
+                hash[:display_name] = (programtype.displayName.get).force_encoding("UTF-8")
+            end
+            #TODO: OS can have multiple loads assigned but honeybee schema has a single load in
```
chriswmackey, line 47: As I mentioned earlier, I don't think there's really anything to do here. The way that you implemented this is good.

lib/from_openstudio/program_type.rb
```diff
@@ -0,0 +1,91 @@
...
+            #programtype
+            unless programtype.people.empty?
+                people_def = programtype.people[0].peopleDefinition
+                # Only translate if people per floor area is specified
+                unless people_def.peopleperSpaceFloorArea.empty?
```
chriswmackey, line 52: If you wanted to put checks for whether schedules exist and are of the correct type, I think you can do it here next to this check.

lib/to_openstudio/load/daylight.rb
```diff
@@ -73,7 +73,7 @@ def to_openstudio(openstudio_model, parent_zone, parent_space)
       end
 
       # assign light output fraction if it exists
-      if @hash[:min_power_input]
+      if @hash[:min_light_output]
```
chriswmackey, line 76: Good bug finding here.

chriswmackey (changes requested):
Thanks for getting this together, @tanushree04 . There are a lot of little bugs and edge cases to address but it's generally looking good. As I said in the comments, I think we can forego the implementation of service hot water for now. Maybe we can find a way to do it later.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,129 @@ def self.properties_from_space(space)
...
+      # set room energy properties
+      unless space.defaultConstructionSet.empty?
+        const_set = space.defaultConstructionSet.get
+        hash[:construction_set] = const_set.nameString
+      end
+      unless space.spaceType.empty?
+        space_type = space.spaceType.get
+        hash[:program_type] = space_type.nameString
+      end
+      # TODO: These are loads assigned to the space directly. How should duplicates created in programtype, if any, be handled?
```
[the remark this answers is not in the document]
tanushree04: I think that makes sense and if we have a way to override the honeybee room assigned loads over the program type we should stick to that.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,129 @@ def self.properties_from_space(space)
...
+      end
+      unless space.spaceType.empty?
+        space_type = space.spaceType.get
+        hash[:program_type] = space_type.nameString
+      end
+      # TODO: These are loads assigned to the space directly. How should duplicates created in programtype, if any, be handled? 
+      unless space.people.nil?
+        space.people.each do |people|
+          unless people.peopleDefinition.peopleperSpaceFloorArea.empty?
+            hash[:people] = Honeybee::PeopleAbridged.from_load(people)
```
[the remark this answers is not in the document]
tanushree04: That's a good suggestion!

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,129 @@ def self.properties_from_space(space)
...
+          electric_eq_def = electric_eq.electricEquipmentDefinition
+          # Only translate if watts per floor area is specified
+          unless electric_eq_def.wattsperSpaceFloorArea.empty?
+            hash[:electric_equipment] = Honeybee::ElectricEquipmentAbridged.from_load(electric_eq)
+          end
+        end
+      end
+      unless space.gasEquipment.nil?
+        space.gasEquipment.each do |gas_eq|
+          hash[:gas_equipment] = Honeybee::GasEquipmentAbridged.from_load(gas_eq)
```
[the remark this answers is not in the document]
tanushree04: good catch

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,129 @@ def self.properties_from_space(space)
...
+        end
+      end
+      #TODO: The translation of other equipment to HB JSON is not working.
+      unless space.otherEquipment.nil?
+        space.otherEquipment.each do |other_eq|
+          hash[:process_loads] = Honeybee::ProcessAbridged.from_load(other_eq)
```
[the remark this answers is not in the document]
tanushree04: That makes sense since users can probably use process_loads to define various kinds of other equipment.

lib/from_openstudio/program_type.rb
```diff
@@ -0,0 +1,125 @@
...
+            end
+            unless programtype.people.empty?
+                programtype.people.each do |people|
+                    people_def = people.peopleDefinition
+                    # Only translate if people per floor area is specified
+                    # Check if schedule exists and is of the correct type
+                    if !people_def.peopleperSpaceFloorArea.empty? && !people.numberofPeopleSchedule.empty?
+                        sch = people.numberofPeopleSchedule.get.scheduleTypeLimits.get.unitType
+                        # TODO Fractional isn't listed in schedule type limits unit type
+                        if sch == 'Dimensionless'
```
tanushree04, line 55: @chriswmackey as per the schema these should be `Fractional` schedules but we havent added a `Fractional` unit type to the enum for schedule type limits

spec/tests/osm_material_spec.rb
joseph-robertson: @tanushree04

(1) Where are you setting the calendar year? Here is a snippet for setting it: https://github.com/NREL/resstock/blob/develop/resources/measures/HPXMLtoOpenStudio/resources/location.rb#L65-L68

(2) Above you are setting the weather file on the osw, and not the model. Here is a snippet for setting it: https://github.com/NREL/resstock/blob/develop/resources/measures/HPXMLtoOpenStudio/resources/location.rb#L37-L45

lib/from_openstudio/geometry/room.rb
tanushree04: Yes that makes sense!

spec/tests/osm_material_spec.rb
tanushree04: @joseph-robertson thanks for sharing this.
After loading my openstudio model 

1) I am setting the calendar year here: https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/251/files#diff-d7d4c83dbf9adb63b5ff4c9e87720d7754ec1dbad33876ad970dc892313455cbR47
Does this look correct? The resultant model does show 2020 as the calendar year

2) I am not setting the weather file to the model : https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/251/files#diff-d7d4c83dbf9adb63b5ff4c9e87720d7754ec1dbad33876ad970dc892313455cbR51

joseph-robertson: @tanushree04 Does your simulation still error out?

tanushree04: Yes, it is still printing out the warnings.
The command I use to run it is `bundle exec rake` which is running all the spec files.

lib/from_openstudio/model.rb
```diff
@@ -281,27 +281,35 @@ def self.schedtypelimits_from_model(openstudio_model)
...
         sched_fixed_hash = ScheduleFixedIntervalAbridged.from_schedule_fixedinterval(sch_fix_int, is_leap_year)
         $schedules[sched_fixed_hash[:identifier]] = sched_fixed_hash
         result << sched_fixed_hash
       end
+      unless $heating_setpoint_schedule.nil?
+        result << $heating_setpoint_schedule
+      end
+      unless $cooling_setpoint_schedule.nil?
+        result << $cooling_setpoint_schedule
+      end
```
chriswmackey, line 303: It would seem that this code is not doing anything right now since this `$heating_setpoint_schedule` variable isn't really created until the Room translation is done, which is happening after the translation of the schedules. Can we just move these 6 lines of code to happen after the room translation in `self.translate_from_openstudio`?

chriswmackey (commented):
I just found a small bug. Am I right that you still are implementing the check for ScheduleRuleset and ScheduleFixedInterval, @tanushree04 ?

FYI, I just tested @joseph-robertson 's [code to import ScheuleYear and ScheduleRuleset](https://github.com/NREL/OpenStudio/pull/4508) and it's working well for shade transmittance schedules. I'm looking forward to trying it with the schedules for the loads once we merge this PR.

lib/from_openstudio/model.rb
tanushree04: Thanks for catching that @chriswmackey.
I have implemented the check for ScheduleRuleset and ScheduleFixedInterval.

spec/tests/osm_construction_spec.rb
```diff
@@ -62,6 +69,10 @@
     file = File.join(File.dirname(__FILE__), '../samples/osm_construction/energyConstructionAirBoundary.osm')
     vt = OpenStudio::OSVersion::VersionTranslator.new
     openstudio_model = vt.loadModel(file)
+    weather_file = File.join(File.dirname(__FILE__), '../samples/epw')
```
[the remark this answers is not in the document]
tanushree04: @joseph-robertson thanks for catching that, I might have missed making the change here.

lib/from_openstudio/load/process.rb
tanushree04: For Process load it isn't. I have added a check for the schedule type here.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,166 @@ def self.properties_from_space(space)
...
+              hash[:gas_equipment] = Honeybee::GasEquipmentAbridged.from_load(gas_eq)
+              break
+            end
+          end
+        end
+      end
+      unless space.otherEquipment.empty?
+        hash[:process_loads] = []
+        space.otherEquipment.each do |other_eq|
+          unless other_eq.designLevel.empty?
```
chriswmackey, line 143: We also need the `sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized` check here.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,166 @@ def self.properties_from_space(space)
...
+              openstudio_sch_type_lim.setNumericType('Temperature')
+              openstudio_schedule.defaultDaySchedule.setName('Heating Day Default')
+              openstudio_schedule.defaultDaySchedule.addValue(OpenStudio::Time.new(0,24,0,0), 100)
+              openstudio_schedule.defaultDaySchedule.setScheduleTypeLimits(openstudio_sch_type_lim)
+              $heating_setpoint_schedule = Honeybee::ScheduleRulesetAbridged.from_schedule_ruleset(openstudio_schedule)
+            end
+            hash[:setpoint][:heating_schedule] = $heating_setpoint_schedule[:identifier]
+          else
+            heating_schedule = thermostat.heatingSetpointTemperatureSchedule.get
+            hash[:setpoint][:heating_schedule] = heating_schedule.nameString
```
chriswmackey, line 192: We need the `sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized` check here.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,166 @@ def self.properties_from_space(space)
...
+              openstudio_sch_type_lim.setNumericType('Temperature')
+              openstudio_schedule.defaultDaySchedule.setName('Cooling Day Default')
+              openstudio_schedule.defaultDaySchedule.addValue(OpenStudio::Time.new(0,24,0,0), -100)
+              openstudio_schedule.defaultDaySchedule.setScheduleTypeLimits(openstudio_sch_type_lim)
+              $cooling_setpoint_schedule = Honeybee::ScheduleRulesetAbridged.from_schedule_ruleset(openstudio_schedule)
+            end
+            hash[:setpoint][:cooling_schedule] = $cooling_setpoint_schedule[:identifier]
+          else
+            cooling_schedule = thermostat.coolingSetpointTemperatureSchedule.get
+            hash[:setpoint][:cooling_schedule] = cooling_schedule.nameString
```
chriswmackey, line 212: We need the `sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized` check here.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,166 @@ def self.properties_from_space(space)
...
+          end
+        end
+        unless thermal_zone.zoneControlHumidistat.empty?
+          humidistat = thermal_zone.zoneControlHumidistat.get
+          unless humidistat.humidifyingRelativeHumiditySetpointSchedule.empty?
+            humidifying_schedule = humidistat.humidifyingRelativeHumiditySetpointSchedule.get
+            hash[:setpoint][:humidifying_schedule] = humidifying_schedule.nameString
```
chriswmackey, line 219: We need the `sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized` check here.

lib/from_openstudio/geometry/room.rb
```diff
@@ -64,12 +64,166 @@ def self.properties_from_space(space)
...
+          end
+          unless humidistat.dehumidifyingRelativeHumiditySetpointSchedule.empty?
+            dehumidifying_schedule = humidistat.dehumidifyingRelativeHumiditySetpointSchedule.get
+            hash[:setpoint][:dehumidifying_schedule] = dehumidifying_schedule.nameString
```
chriswmackey, line 223: We need the `sch.to_ScheduleFixedInterval.is_initialized or sch.to_ScheduleRuleset.is_initialized` check here.

lib/from_openstudio/load/electric_equipment.rb
```diff
@@ -0,0 +1,66 @@
...
+            hash[:type] = 'ElectricEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
+                    hash[:schedule] = schedule.nameString
```
chriswmackey, line 50: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/gas_equipment.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        def self.from_load(load)
+            # create an empty hash
+            hash = {}
+            hash[:type] = 'GasEquipmentAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 46: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/infiltration.rb
```diff
@@ -0,0 +1,62 @@
...
+            hash[:type] = 'InfiltrationaAridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            hash[:flow_per_exterior_area] = load.flowperExteriorSurfaceArea.get
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 50: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/lighting.rb
```diff
@@ -0,0 +1,66 @@
...
+            hash = {}
+            hash[:type] = 'LightingAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            unless load.schedule.empty?
+                schedule = load.schedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 49: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/people.rb
```diff
@@ -0,0 +1,75 @@
...
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            unless load.displayName.empty?
+                hash[:display_name] = (load.displayName.get).force_encoding("UTF-8")
+            end
+            #TODO: Is there a default occupancy schedule since it is a required field in HB but not
+            #in OS
+            unless load.numberofPeopleSchedule.empty?
+                schedule = load.numberofPeopleSchedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 51: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/people.rb
```diff
@@ -0,0 +1,75 @@
...
+                    hash[:occupancy_schedule] = schedule.nameString
+                end
+            end
+            unless load.activityLevelSchedule.empty?
+                schedule = load.activityLevelSchedule.get
+                if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 57: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/process.rb
```diff
@@ -0,0 +1,63 @@
...
+            hash[:type] = 'ProcessAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            load_def = load.otherEquipmentDefinition
+            unless load_def.displayName.empty?
+                hash[:display_name] = (load_def.displayName.get).force_encoding("UTF-8")
+            end
+            hash[:watts] = load_def.getDesignLevel
+            schedule = load.schedule
+            if schedule.to_ScheduleFixedInterval.is_initialized or schedule.to_ScheduleRuleset.is_initialized
```
chriswmackey, line 50: We should not need this check here since we perform it before we make the load object

lib/from_openstudio/load/ventilation.rb
```diff
@@ -0,0 +1,60 @@
...
+            hash[:identifier] = clean_name(load.nameString)
+            hash[:air_changes_per_hour] = load.outdoorAirFlowAirChangesperHour
+            hash[:flow_per_zone] = load.outdoorAirFlowRate
+            hash[:flow_per_person] = load.outdoorAirFlowperPerson
+            hash[:flow_per_area] = load.outdoorAirFlowperFloorArea
+            unless load.outdoorAirFlowRateFractionSchedule.empty?
+                sch = load.outdoorAirFlowRateFractionSchedule.get.scheduleTypeLimits.get.unitType
+                # Check if the schedule type is correct
+                if sch == 'Fractional'
+                    hash[:schedule] = load.name
```
chriswmackey, line 52: This seems a little odd. Shouldn't this be `load.outdoorAirFlowRateFractionSchedule.get.nameString`? Also, it seems ok if the unit type isn't Fractional as long as the values are between 0 and 1 so I think we can remove that check.

lib/to_openstudio/model.rb
```diff
@@ -89,6 +89,7 @@ def to_openstudio_model(openstudio_model=nil, log_report=true)
                             OpenStudio::Model::Model.new
                           end
 
+      @openstudio_model.getYearDescription.setCalendarYear(2020)
```
chriswmackey, line 92: We should probably set this to 2017 since the Ladybug Tools default is to have Jan 1 be on a Sunday.

lib/to_openstudio/simulation/parameter_model.rb
```diff
@@ -49,6 +49,8 @@ def to_openstudio_model(openstudio_model=nil, log_report=false)
                           else
                             OpenStudio::Model::Model.new
                           end
+      
+      @openstudio_model.getYearDescription.setCalendarYear(2020)
```
chriswmackey, line 53: This should also be set to 2017.

lib/to_openstudio/model.rb
chriswmackey: Or it should be 2016 if the EPW is for a leap year. Hopefully, this will just correct itself when the EPW gets assigned.

spec/tests/honeybee_construction_set_spec.rb
```diff
@@ -35,6 +35,7 @@
   # TODO: add assertions about properties
   it 'can load construction set complete' do
     openstudio_model = OpenStudio::Model::Model.new
+    openstudio_model.getYearDescription.setCalendarYear(2020)
```
chriswmackey, line 38: Ah, we have a lot of these to update across the tests.

spec/tests/honeybee_schedule_spec.rb
```diff
@@ -182,8 +192,8 @@
     expect(filename).to eq 'Solstice_Increasing.csv'
     columns = schedule_csv[:columns]
     expect(columns.size).to eq 2
-    expect(columns[0].size).to eq 8760*6 + 1
-    expect(columns[1].size).to eq 8760*6 + 1
+    expect(columns[0].size).to eq 8784*6 + 1
+    expect(columns[1].size).to eq 8784*6 + 1
```
chriswmackey, line 196: Hm. Come to think of it, why did we pick a leap year? This is definitely the rarer case. I'll change this back once we set the default calendar year to 2017.

chriswmackey (approved):
This is looking good. I have several comments but they are all minor so I think I'll just merge this in and push a fix for them shortly.

Thanks again for all of this, @tanushree04 !

PR #277, opened 2022-03-17
feat(ServiceHotWater): Adds Service Hot water reverse translator

### LINKED ISSUE #264: Import Hot Water Loads from WaterUseEquipment and WaterUseEquipmentDefinition
We know it's too complex to import detailed SHW systems or some of the plant side properties like the water temperature but it would be really nice if we could import just the usage `schedule` and the `flow_per_area` to the [ServiceHotWaterAbridged](https://www.ladybug.tools/honeybee-schema/model.html#tag/servicehotwaterabridged_model) object.  We should be able to take these from the [WaterUseEquipment](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.3.0-doc/model/html/classopenstudio_1_1model_1_1_water_use_equipment.html) objects and [WaterUseEquipmentDefinition](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.3.0-doc/model/html/classopenstudio_1_1model_1_1_water_use_equipment_definition.html) objects in the model.

tanushree04:
Fixes #264 

- Adds reverse translator for Service Hot Water
- At the room level, creates new reverse translator object if OS Water Use Equipment object is present

### DIFF
--- a/lib/from_openstudio/load/service_hot_water.rb
+++ b/lib/from_openstudio/load/service_hot_water.rb
@@ -1,0 +1,60 @@
+# *******************************************************************************
+# Honeybee OpenStudio Gem, Copyright (c) 2020, Alliance for Sustainable
+# Energy, LLC, Ladybug Tools LLC and other contributors. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# (1) Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# (2) Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# (3) Neither the name of the copyright holder nor the names of any contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission from the respective party.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDER(S) AND ANY CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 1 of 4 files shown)


lib/from_openstudio/load/service_hot_water.rb
```diff
@@ -0,0 +1,56 @@
...
+            hash[:type] = 'ServiceHotWaterAbridged'
+            # set hash values from OpenStudio Object
+            hash[:identifier] = clean_name(load.nameString)
+            load_def = load.WaterUseEquipmentDefinition
+            # units of peak flow rate are m3/s
+            peak_flow = load_def.peakFlowRate
+            # unit for flow per area is L/h-m2 (m3/s = 3600000 L/h)
+            hash[:flow_per_area] = (peak_flow * 3600000)/floor_area
+            unless load_def.flowRateFractionSchedule.empty?
+                hash[:schedule] = load_def.flowRateFractionSchedule.get.nameString
```
chriswmackey, line 50: What about the case that this schedule is not a ScheduleRuleset or ScheduleFixedInterval? I think you need something like this before you call this `from_load` method:
```
if sch.to_ScheduleRuleset.is_initialized or sch.to_ScheduleFixedInterval.is_initialized
```

chriswmackey (commented):
This generally looks good, @tanushree04 . I think we just need a check to make sure the class of the schedule is Ruleset or FixedInterval. Then we should be able to merge.

lib/from_openstudio/load/service_hot_water.rb
tanushree04: @chriswmackey thanks for catching that, as per the schema this should be a fractional schedule. Does it also accept schedule ruleset?

chriswmackey: Yes.

chriswmackey (approved):
Let's merge!

PR #284, opened 2022-05-09
feat(hvac): Add ideal air system

### LINKED ISSUE #263: Consider Importing Ideal Air Systems from OpenStudio to HBJSON
We know that trying to import detailed HVAC systems is too complex to be worthwhile but Ideal Air Systems are simple enough that we should be able to translate them from OSM to HBJSON if we find them. There are only [a handful of properties that we would want to translate](https://www.ladybug.tools/honeybee-schema/model.html#tag/idealairsystemabridged_model) so this hopefully should not take long to implement. As long as we find a way to determine whether the zone equipment assigned to a given ThermalZone is an Ideal Air System so that we don't get errors when detailed systems are assigned.

tanushree04:
Fixes #263 

TODO: Add test

### DIFF
--- a/lib/from_openstudio/geometry/room.rb
+++ b/lib/from_openstudio/geometry/room.rb
@@ -201,6 +201,14 @@ module Honeybee
       thermal_zone = space.thermalZone
       unless thermal_zone.empty?
         thermal_zone = space.thermalZone.get
+        # Create ideal_air_system if present
+        unless thermal_zone.equipment.nil?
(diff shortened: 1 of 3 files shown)


chriswmackey:
Glad to see this coming together. I'll review it after the tests are added.

chriswmackey (approved):
I think I'll merge this just so that I can do some integration tests on my end before the next stable release of the LBT plugin.

We can always add the unit tests later.

